PR #1, opened 2020-06-23
Add gitignore

lukasberglund:
Git will now ignore files that are generated by Maven during a build.


lukasberglund:
@googlebot I signed it!

PR #2, opened 2020-06-23
Initialize Maven

lukasberglund:
Initialize maven project that with dummy tests.


capstone/src/main/java/war/App.java
```diff
@@ -0,0 +1,13 @@
+package war;
```
winniewan, line 1: Why is it named war? 
Is it possible to follow the package standard names from tutorial?

capstone/src/main/java/war/App.java
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Hello world!
+ *
+ */
+public class App 
+{
```
winniewan, line 8: Can you remove the line break before the opening brace? 
Here's the Java style guideline: K & R style 
https://google.github.io/styleguide/javaguide.html#s4.1.2-blocks-k-r-style

capstone/src/main/java/war/App.java
```diff
@@ -0,0 +1,13 @@
...
+    public static void main( String[] args )
```
winniewan, line 9: Can you remove the space after opening and before ending parentheses? Like this: main(String[] args)
Java style guidelines for spaces: https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace

lukasberglund: Yeah this was all automatically generated by Maven. I'll change it.

capstone/src/main/java/war/App.java
lukasberglund: I changed it to that, my bad

capstone/src/main/java/com/google/sps/App.java
```diff
@@ -0,0 +1,25 @@
...
+// limitations under the License.
+
+package com.google.sps;
+
+/**
+ * Hello world!
+ *
+ */
+public class App {
+    public static void main(String[] args) {
```
winniewan, line 22: Thanks for addressing the nits before! 
Just one more nit: any indentation should be +2 spaces instead of a tab :)  
https://google.github.io/styleguide/javaguide.html#s4.2-block-indentation

lukasberglund: Ah shoot.

PR #3, opened 2020-06-24
Create profile prototype

winniewan:
There are two types of files to demonstrate how the profile page will look like. 
There is a view and edit option for profile page. 

The home page is to view businesses profiles as index cards.


capstone/src/main/webapp/editProfile.html
```diff
@@ -0,0 +1,55 @@
...
+  </head>
+  <body>
+    <nav>
+      <div class="logo">
+        <h1>Support Local Businesses</h1>
+      </div>
+      <ul class="nav">
+        <li><a href="index.html">Home</a></li>
+        <li><a href="viewProfile.html">View Profile</a></li>
+        <li><a href="editProfile.html">Edit Profile</a></li>
```
eash130: Should these be separate or do you think we could add some sort of edit button on the profile page to make changes?

capstone/src/main/webapp/editProfile.html
```diff
@@ -0,0 +1,55 @@
...
+            I would head off to the nearby beach and take a swim. Painting is a little
+            difficult for me, but I'm getting there! 
+          </textarea>
+          <input type="submit" value="Submit">
+        </form> 
+      </div>
+    </div>
+
+    <div class="footer">
+      <p>If you have questions, click on this link to contact us! :)</p>
```
eash130, line 51: What will the link be? Email link, google form, etc.?

capstone/src/main/webapp/editProfile.html
winniewan: Check the new commit :)

eash130: I see what you did in the recent commit. To elaborate on my comment, I was thinking of a single file profile.html that changes between input tags and p tags in the javascript depending on if you are editing that field or not.

capstone/src/main/webapp/editProfile.html
winniewan: I just hard coded something for prototype. 
We can discuss what the footer consist of in a meeting!

capstone/src/main/webapp/editProfile.html
winniewan: That's a great suggestion! I was thinking of having two files because there will be a lot of additionally features added to the profile page. If we combine it to one and toggle it with JS, it might be a large HTML file. What do you think about this?

winniewan: Discussed over GVC and we both agree to implement it as one HTML file to reduce redundancy!

lukasberglund:
Some screenshots for context: 
https://screenshot.googleplex.com/OkrKJmy2Km4
https://screenshot.googleplex.com/S0ZQZCDSoUi
https://screenshot.googleplex.com/eNhg9Xjcnoz

alygerharter:
> Some screenshots for context:
> https://screenshot.googleplex.com/OkrKJmy2Km4
> https://screenshot.googleplex.com/S0ZQZCDSoUi
> https://screenshot.googleplex.com/eNhg9Xjcnoz

PR #4, opened 2020-06-24
Add dependency for appengine

lukasberglund:
This allows us to run the command `mvn package appengine:run`

### DIFF
--- a/capstone/pom.xml
+++ b/capstone/pom.xml
@@ -25,6 +25,11 @@
       <version>4.11</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.google.appengine</groupId>
+      <artifactId>appengine-api-1.0-sdk</artifactId>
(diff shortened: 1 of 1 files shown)


capstone/pom.xml
```diff
@@ -69,6 +74,15 @@
           <artifactId>maven-project-info-reports-plugin</artifactId>
           <version>3.0.0</version>
         </plugin>
+        <plugin>
+          <groupId>com.google.cloud.tools</groupId>
+          <artifactId>appengine-maven-plugin</artifactId>
+          <version>2.2.0</version>
+          <configuration>
+            <deploy.projectId>bergmoney-step-2020</deploy.projectId>
```
winniewan, line 82: Do we have a project id for our capstone project?

lukasberglund: Ohh. Yes, probably. I don't know it though.

PR #5, opened 2020-06-24
Rewrite pom.xml based on file used in training.

lukasberglund:
This should fix issues with running the server locally.


capstone/pom.xml
```diff
@@ -69,6 +74,15 @@
           <artifactId>maven-project-info-reports-plugin</artifactId>
           <version>3.0.0</version>
         </plugin>
+        <plugin>
+          <groupId>com.google.cloud.tools</groupId>
+          <artifactId>appengine-maven-plugin</artifactId>
+          <version>2.2.0</version>
+          <configuration>
+            <deploy.projectId>bergmoney-step-2020</deploy.projectId>
```
winniewan: If we have the project id, maybe update this.

eash130: I believe that it might be cloud-build-step-2020, but you can confirm with Ben if this is the final project ID

capstone/pom.xml
```diff
@@ -1,75 +1,62 @@
...
+        <version>2.2.0</version>
+        <configuration>
+          <deploy.projectId>cloud-build-step-2020</deploy.projectId>
+          <deploy.version>1</deploy.version>
+        </configuration>
+      </plugin>
+    </plugins>
   </build>
-</project>
+</project>
```
alygerharter, line 62: please add a newline at the end of the file

lukasberglund: I always forget that!

PR #7, opened 2020-06-24
Add business prototype page

winniewan:


lukasberglund:
Can you post a screenshot of this in the group?

capstone/src/main/webapp/business.html
```diff
@@ -0,0 +1,86 @@
...
+      </div class="comment-section">
+        <!-- TODO: Add comment section here. -->
+      </div>
+    </div>
+
+    <div class="footer">
+      <p>If you have questions, click on this link to contact us! :)</p>
+    </div>
+  </body>
+</html>
```
lukasberglund, line 86: All this looks pretty good to me :) One thing to mention is that eventually you are going to have to write javascript code that generates this based on data provided. You could do it during MVP stages or now I guess.

capstone/src/main/webapp/js/profile.js
```diff
@@ -3,11 +3,11 @@ function showProfile() {
   var viewProfile = document.getElementById("view-profile-section");
```
lukasberglund, line 3: Can we call this function something more accurate like `toggleProfile`?

capstone/src/main/webapp/business.html
```diff
@@ -0,0 +1,86 @@
...
+        
+        <h3>How You Can Help Us</h3>
+
+        <p>
+          It would mean the world if you can support us during this time!
+          Here is how you can help:
+          <ol>
+            <li>Click here to donate</li>
+            <li>Click here to order from us</li>
+            <li>Click here to share our story</li>
```
lukasberglund, line 71: How are you gonna get these links?

capstone/src/main/webapp/js/profile.js
winniewan: Yep, done!

capstone/src/main/webapp/business.html
winniewan: The business should provide these links. We can give them suggestions on what they can provide or share if they don't have any ideas on top of mind. We can include additional functionalities later on that may help them with this.

capstone/src/main/webapp/business.html
alygerharter: hard coded data is fine for now, i think. but good to call out that this will need to change eventually!

PR #12, opened 2020-06-26
Display replies on comments

lukasberglund:
Display replies on comments. Replies can be accessed by clicking the show reply button on individual comments.


capstone/src/main/webapp/js/util.js
```diff
@@ -15,4 +15,12 @@
...
   return new Promise((resolve, reject) => resolve(val));
+}
+
+/** Build html element of specified type and content */
+export function buildElement(type, content) {
+  let element = document.createElement(type);
+  element.innerText = content;
+
+  return element;
 }
```
winniewan, line 26: Add a line to end of file

capstone/src/main/webapp/js/comments.js
```diff
@@ -51,12 +51,14 @@ function getUserName(userId) {
...
-function buildElement(type, content) {
-  let element = document.createElement(type, content);
-  element.innerText = content;
-
-  return element;
+function getReplies(commentId) {
+  // TODO (bergmoney@): Request replies from API
+  let hash = commentId + 10;
+  let replies = [
+    {id: 0 + hash, userId: 0, content: 'Hii' + hash, timestamp: 1},
```
alygerharter, line 58: did you mean to add the hash to the content?

lukasberglund: I think I did, but I don't remember why exactly lol.

alygerharter: from your screenshot, it looks like the comment says "Hii11" which i don't think is intended :)

PR #13, opened 2020-06-26
Display timestamps on comments

lukasberglund:
This allows users to see how long ago a comment was posted. Depending on how long ago the comment was posted, either the seconds, minutes, hours, days, weeks, months, or years since the comment was posted are displayed at the top of the comment.


capstone/src/main/webapp/js/util.js
```diff
@@ -17,6 +17,66 @@ export function wrapInPromise(val) {
...
+  
+  // Could be negative (e.g. when comparing January 2020 with December 2019)
+  let monthDiff = now.getMonth() - then.getMonth();
+
+  return yearDiff * 12 + monthDiff;
+}
+
+/** 
+* Format timestamp by showing the minutes/hours/days/weeks/months/years passed since date */
+export function showTimeElapsedSince(then) {
```
prabenzom, line 41: Are you able to use an existing library to do this? I think that would be best. If not, you should write unit tests to test this function.

eash130: I think implementing the timestamp feature might be easier when we add the servlet backends since you can store the time along with the comment.

lukasberglund: Yeah that's probably right.

prabenzom: Even if we store the time along with the comment, won't we still have the same problem to solve client-side? The backend will just store a date, but it seems like in the front end we may want to show 10 minutes ago, 1 hour ago, etc. Either way i'm fine to punt this until later.

lukasberglund: I was thinking we return another field from the servlet that says how long ago the comment was posted.

PR #17, opened 2020-06-26
Convert home.* to index.* and add additional businesses to the page.

eash130:
Renamed:
home.html -> index.html
home.css -> index.css
home.js -> index.js

Implemented an onload of fetching businesses.


capstone/src/main/webapp/index.html
```diff
@@ -14,12 +14,10 @@ <h1>Support Local Businesses</h1>
       <ul class="nav">
         <li><a href="index.html">Home</a></li>
         <li><a href="profile.html">Profile</a></li>
-        <li id="logout-button">Log out</li>
+        <li id="authentication-button">Log in</li>
```
winniewan, line 17: We should probably keep the authentication consistent throughout the prototype. Is it possible to keep it as "Log out" because I have the user's profile page with edit button?

eash130: I'll just revert it back to what it was. I made this change a while back.

capstone/src/main/webapp/js/index.js
```diff
@@ -0,0 +1,73 @@
...
+    businessLink.appendChild(buildElement('h2', getBusinessNames()[i]));
+    businessCard.appendChild(businessLink);
+
+    // Add description to business.
+    let businessInfo = buildElement('p', getBusinessDescriptions()[i]);
+    businessInfo.classList.add('business-info');
+    businessCard.appendChild(businessInfo);
+
+    // Add show more button.
+    let showMore = buildElement('p', '');
```
winniewan, line 22: Is it possible to make helper functions?

eash130: I think that's a good idea to do.

eash130:
@googlebot I signed it!

PR #18, opened 2020-06-26
Add comment section to profile and business page

lukasberglund:
Generate comments fully via javascript thereby removing `comments.html` export function for loading comments and use it in `businesses.js` and `profile.js`


capstone/src/main/webapp/js/business.js
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an 'AS IS' BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { loadCommentSection } from '/js/comments.js';
+
+window.onload = () => loadCommentSection(document.getElementById('comment-section'));
```
alygerharter, line 17: nit: add a newline

capstone/src/main/webapp/js/comments.js
```diff
@@ -17,10 +17,37 @@ import { wrapInPromise, buildElement } from '/js/util.js';
 var commentField = undefined;
 var commentContainer = undefined;
 
-window.onload = () =>{
-  commentField = document.getElementById('comment-field');
+window.onload = () => loadCommentSection(document.getElementById('comment-section'));
```
alygerharter, line 20: do you need this line? since you have it in both business.js and profile.js

lukasberglund: Well I am loading comments on both of those pages, so I need the line in the sense that I need the logic to be executed twice.

capstone/src/main/webapp/js/business.js
lukasberglund: It's already there :)

capstone/src/main/webapp/js/comments.js
alygerharter: what i mean is you have this exact line of code in business.js, profile.js, and comments.js. does it need to be in all 3 places?

lukasberglund: Ohhh. Wait you are right. I am getting rid of it.

PR #19, opened 2020-06-29
Add authentication login/logout buttons to webpage.

winniewan:
Use Users API to add authentication. 

Add login/logout buttons.
If a user is a new user, they will be redirected to a "Are you a business" form. 
If a user is a returning user, they will be redirected to home page.


capstone/src/main/java/com/google/sps/servlets/NewUserServlet.java
```diff
@@ -0,0 +1,51 @@
...
+
+    try {
+      Entity ent = datastore.get(userKey);
+    } catch (EntityNotFoundException e) {
+      // Store new user into database.
+      Entity userEntity = new Entity("User", userId);
+      datastore.put(userEntity);
+      
+      // Redirect to the questionnaire page.
+      response.sendRedirect("/questionnaire.html");
```
prabenzom, line 44: can a user just navigate away from this page back to index and not fill out the questionnaire?  It sounds like we should perhaps just have a sign up flow separate from the auth flow? i.e. 

Sign-up:  get logged in user, see if user is in the database. 
 -If yes, they should log in (or just log them in? how do other websites generally handle google login?).  
- if not, add them to the database only after filling out the questionnaire. If they navigate away from the page, they are not logged in, have to do sign up flow again. 

log-in - check if user is in database, if so, they are now logged in.

capstone/src/main/java/com/google/sps/servlets/AuthenticateServlet.java
```diff
@@ -0,0 +1,40 @@
...
+/**
+*  Authentication for the webpage. Uses Users API for authentication functionality. 
+*  Create login/logout url links for client side.
+*  When signed in, redirect to a servlet to determine whether it's a new user or not. 
+*/
+@WebServlet("/auth")
+public class AuthenticateServlet extends HttpServlet {
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
prabenzom, line 22: please add tests for this method. We should test that we get the appropriate json for logged in vs not logged in users

capstone/src/main/java/com/google/sps/servlets/NewUserServlet.java
```diff
@@ -0,0 +1,51 @@
...
+*  Check if user is a new user signing in or not.
+*  If it's a new user, redirect to questionnaire page to determine whether they're a 
+*  business owner. 
+*  Add new user to database.
+*/
+@WebServlet("/check_new_user")
+public class NewUserServlet extends HttpServlet {
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
prabenzom, line 28: Please add tests to this after we confirm what the flow should look like

capstone/src/main/java/com/google/sps/servlets/AuthenticateServlet.java
```diff
@@ -0,0 +1,40 @@
...
+    UserService userService = UserServiceFactory.getUserService();
+    User userData;
+
+    // If user is logged in, redirect to home page. 
+    // Otherwise, redirect to NewUserServlet to determine whether it's a new user signing in or not.
+    if (userService.isUserLoggedIn()) {
+      String logoutUrl = userService.createLogoutURL("/index.html");
```
prabenzom, line 29: one thing to note... this logs you out of your google account, not just your website. I'd add a warning indicating this (i got logged out of my @google.com) address. Something you may want to consider doing is adding a layer ontop of user service. https://stackoverflow.com/questions/14177469/signing-out-appengine-app-without-signing-out-from-google-com/14178401#14178401

PR #21, opened 2020-06-30
Revert "Add Java lint and ESLint"

winniewan:
Reason: ESLint is not found in github config


lukasberglund:
Could you elaborate on what the reason is?

PR #23, opened 2020-06-30
Implement BusinessServlet

eash130:
Unit tests coming in a future commit, I just want some feedback on the servlet before I write the tests.


capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -0,0 +1,84 @@
...
+  private static final String BUSINESS_NAME = "businessName";
+  private static final String NAME_PROPERTY = "name";
+  private static final String EMAIL_PROPERTY = "email";
+  private static final String BIO_PROPERTY = "bio";
+  private static final String LOCATION_PROPERTY = "location";
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // If the user is navigating to a certain business, then only display that information.
+    String businessName = request.getParameter(BUSINESS_NAME);
```
prabenzom, line 47: a get for a specific business should be mapped to /business/{BUSINESS_NAME}. or /business/{BUSINESS_ID}. I think Id is better because it's unique. multiple businesses could have the same name. This would be a separate servlet. The logic for Listing businesses should be in "/businesses".

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -0,0 +1,84 @@
...
+        BusinessProfile business = new BusinessProfile(name, "", bio, "");
+        businesses.add(business);
+      }
+      Gson gson = new Gson();
+      String jsonBusinesses = gson.toJson(businesses);
+    } else {
+      // Retrieve all of the information for a single business to be displayed.
+      Query businessQuery =
+          new Query(TASK_NAME)
+              .setFilter(new FilterPredicate(NAME_PROPERTY, FilterOperator.EQUAL, businessName));
```
prabenzom, line 68: what if name isn't unique?. You *could* add a parameter  for filtering, but given the current properties of a business, i'm not sure what you could filter on. (for example, you could filter on business types)

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -0,0 +1,84 @@
...
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
+import com.google.gson.Gson;
+import com.google.sps.data.BusinessProfile;
+import java.io.IOException;
+import java.util.ArrayList;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
```
lukasberglund, line 31: Have you made sure all these imports are required?

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -0,0 +1,84 @@
...
+      Query businessQuery = new Query(TASK_NAME);
+      DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+      PreparedQuery queryResults = datastore.prepare(businessQuery);
+      ArrayList<BusinessProfile> businesses = new ArrayList();
+      for (Entity businessEntity : queryResults.asIterable()) {
+        String name = (String) businessEntity.getProperty(NAME_PROPERTY);
+        String bio = (String) businessEntity.getProperty(BIO_PROPERTY);
+        // Create a BusinessProfile of this business with just the relevant information.
+        BusinessProfile business = new BusinessProfile(name, "", bio, "");
+        businesses.add(business);
```
lukasberglund, line 60: Why not add email and location? The Webapp could end up needing it if we add new features after all.

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -0,0 +1,84 @@
...
+      DataStoreService datastore = DatastoreServiceFactory.getDatastoreService();
+      PreparedQuery queryResults = datastore.prepare(businessQuery);
+      Entity businessEntity = queryResults.asSingleEntity();
+      if (businessEntity != null) {
+        String name = (String) businessEntity.getProperty(NAME_PROPERTY);
+        String email = (String) businessEntity.getProperty(EMAIL_PROPERTY);
+        String bio = (String) businessEntity.getProperty(BIO_PROPERTY);
+        String location = (String) businessEntity.getProperty(LOCATION_PROPERTY);
+        BusinessProfile business = new BusinessProfile(name, email, bio, location);
+        String jsonBusinesses = gson.toJson(business);
```
lukasberglund, line 78: Has gson been declared in this block yet? I think it would make more sense to remove this line and then add lines 62-63 after line 80.

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
eash130: Yes, all of the imports are being used.

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
eash130: I didn't add the email and location because this part of the code is mainly for the home page where multiple businesses will be displayed. The home page only needs the name and bio. All of the other info would be accessed when navigating to a single business page. I don't see the harm of adding the email and location though.

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
eash130: Good Catch. I can't move lines 62-63 after line 80 in case of potential scoping errors that may arise with the individual businesses.

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
lukasberglund: Yeah I think it would be good to add.

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
lukasberglund: Ah true

capstone/src/main/java/com/google/sps/data/BusinessProfile.java
```diff
@@ -0,0 +1,16 @@
+package com.google.sps.data;
+
+public final class BusinessProfile {
+
+  private final String name;
```
prabenzom, line 5: id (the one generated by datastore) should be returned imo

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -0,0 +1,68 @@
...
+  private static final String NAME_PROPERTY = "name";
+  private static final String EMAIL_PROPERTY = "email";
+  private static final String BIO_PROPERTY = "bio";
+  private static final String LOCATION_PROPERTY = "location";
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Retrieve the business ID that is located in place of the * in the URL.
+    // request.getPathInfo() returns "/{id}" and substring(1) would return "{id}" without "/".
+    String businessName = request.getPathInfo().substring(1);
```
prabenzom, line 47: what happens if a business name isn't unique? i think this should be id

eash130: I agree. I forgot to actually make the change.

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -0,0 +1,74 @@
...
+/** Servlet that handles business information. */
+@WebServlet("/business/*")
+public class BusinessServlet extends HttpServlet {
+
+  private static final String TASK_NAME = "Business";
+  private static final String BUSINESS_NAME = "businessName";
+  private static final String NAME_PROPERTY = "name";
+  private static final String EMAIL_PROPERTY = "email";
+  private static final String BIO_PROPERTY = "bio";
+  private static final String LOCATION_PROPERTY = "location";
```
lukasberglund, line 42: Nice use of static variables. Good good.

capstone/src/main/java/com/google/sps/servlets/BusinessesServlet.java
```diff
@@ -0,0 +1,66 @@
...
+    ArrayList<BusinessProfile> businesses = new ArrayList();
+
+    // Construct list of businesses from datastore.
+    for (Entity businessEntity : queryResults.asIterable()) {
+      long id = (long) businessEntity.getKey().getId();
+      String name = (String) businessEntity.getProperty(NAME_PROPERTY);
+      String email = (String) businessEntity.getProperty(EMAIL_PROPERTY);
+      String bio = (String) businessEntity.getProperty(BIO_PROPERTY);
+      String location = (String) businessEntity.getProperty(LOCATION_PROPERTY);
+      BusinessProfile business = new BusinessProfile(id, name, email, bio, location);
```
lukasberglund, line 57: This is essentially the same code used to create a business from a datastore entity used in the other file. Is there a way to create a method and then use that method twice? Maybe you can write a method in the `BusinessProfile`. @prabenzom do you know how this is best handled?

prabenzom: a BusinessDatastore class could hold the logic for writing, reading, and updating the business entity in the datastore

capstone/src/main/java/com/google/sps/servlets/BusinessesServlet.java
```diff
@@ -0,0 +1,66 @@
...
+
+  private static final String TASK_NAME = "Business";
+  private static final String BUSINESS_NAME = "businessName";
+  private static final String NAME_PROPERTY = "name";
+  private static final String EMAIL_PROPERTY = "email";
+  private static final String BIO_PROPERTY = "bio";
+  private static final String LOCATION_PROPERTY = "location";
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
prabenzom, line 43: This should be in a follow-up cl, but you should consider what properties we want to filter on going forward for the home page and whether or not they should be added as properties of the business. i.e. what businesses are we showing to non-logged in users (do we just sort by created time, last update time?) what businesses are we showing to logging in users?

eash130:
Unit test for BusinessServlet is on the way. I just wanted some opinions on the way I did BusinessesServletTest

capstone/src/test/java/com/google/sps/servlets/BusinessesServletTest.java
```diff
@@ -0,0 +1,112 @@
...
+    // servletResponse returns "id" from the BusinessProfile
+    business1.setProperty("id", business1.getKey().getId());
+    businesses.add(business1.getProperties());
+
+    Entity business2 = new Entity("Business");
+    business2.setProperty("name", "Business 2");
+    business2.setProperty("email", "work@b2.com");
+    business2.setProperty("bio", "This is a bit about our business");
+    business2.setProperty("location", "New York City, NY");
+    datastore.put(business2);
```
lukasberglund, line 97: I would create a `buildBusinessEntity()` function here to build an entity on one line and save space.

capstone/src/test/java/com/google/sps/servlets/BusinessesServletTest.java
```diff
@@ -0,0 +1,112 @@
...
+    servlet.doGet(request, response);
+    Assert.assertEquals(servletResponseWriter.toString().replace("\n", ""), "[]");
+  }
+
+  @Test
+  public void testBasicdoGet() throws IOException {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+    // This list will help in constructing the expected response.
+    List<Map<String, Object>> businesses = new ArrayList();
```
lukasberglund, line 78: What if you created a list of `Business` objects instead? I think that is going to be a more robust definition of the behavior we expect.

capstone/src/test/java/com/google/sps/servlets/BusinessServletTest.java
```diff
@@ -0,0 +1,127 @@
...
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Test
+  public void testEmptyDatastoredoGet() throws IOException {
+    doReturn("/1").when(request).getPathInfo();
+    servlet.doGet(request, response);
+    // response = "\n" before replacing \n
+    Assert.assertTrue(servletResponseWriter.toString().replace("\n", "").isEmpty());
```
prabenzom, line 69: I think this should return a not found status code if the requested business does not exist. https://www.javamex.com/tutorials/servlets/http_status_code.shtml 

https://stackoverflow.com/questions/35767502/how-to-set-http-status-code-when-responding-to-servlet-client-from-filter-class

capstone/src/test/java/com/google/sps/servlets/BusinessServletTest.java
```diff
@@ -0,0 +1,127 @@
...
+    business1.setProperty("location", "Mountain View, CA");
+    datastore.put(business1);
+
+    // Try to search for a business using an invalid/unregistered ID.
+    doReturn("/" + (business1.getKey().getId() + 1)).when(request).getPathInfo();
+
+    servlet.doGet(request, response);
+    // servletResponse = "\n" before replacing \n.
+    String servletResponse = servletResponseWriter.toString().replace("\n", "");
+    Assert.assertTrue(servletResponse.isEmpty());
```
prabenzom, line 125: same as the first test, this should return a not found error

capstone/src/test/java/com/google/sps/servlets/BusinessesServletTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Test
+  public void testEmptydoGet() throws IOException {
+    servlet.doGet(request, response);
+    Assert.assertEquals(servletResponseWriter.toString().replace("\n", ""), "[]");
```
prabenzom, line 70: I think this is fine to return an empty array (since the client is not asking for a specific entity)

capstone/src/test/java/com/google/sps/servlets/BusinessesServletTest.java
eash130: Good idea. That will save quite a bit of space.

capstone/src/test/java/com/google/sps/servlets/BusinessesServletTest.java
prabenzom: I think passing around business objects does make sense internally (for example, I could imagine a BusinessDatastore object that is responsible for converting between Business and the business entities, but code in servlets just interacts with the Business class).

For this test however, i think we want to verify that the client gets back the expected values - properly formatted Business JSON.

eash130: @prabenzom I saw what you said as well about using BusinessProfile objects and I agree with you both that it could be better than the current implementation. Should I create BusinessProfile objects for each business?

prabenzom: Hm, what do you mean by each business? i'd expect a BusinessProfile object to map to a BusinessEntity

eash130: I meant each business I add into the mock datastore would get a businessprofile to be checked against.

capstone/src/test/java/com/google/sps/servlets/BusinessesServletTest.java
eash130: It returns as a json string, but since there is nothing in it, it sets it to "[]". If I try to return an empty array, it would be an extra step that I'm not sure we need to do.(Assuming I understood the comment correctly)

capstone/src/test/java/com/google/sps/servlets/BusinessServletTest.java
```diff
@@ -0,0 +1,124 @@
...
+    newBusiness.setProperty("location", "Mountain View, CA");
+    return newBusiness;
+  }
+
+  @Test
+  public void testEmptyDatastoredoGet() throws IOException {
+    doReturn("/1").when(request).getPathInfo();
+    servlet.doGet(request, response);
+    // response status is not actually set with mocks so you must check if sendError is called.
+    Mockito.verify(response, Mockito.times(1)).sendError(Mockito.anyInt(), Mockito.anyString());
```
prabenzom, line 79: can you make sure it's a not found error code?

capstone/src/test/java/com/google/sps/servlets/BusinessesServletTest.java
prabenzom: Got it, thanks.

PR #27, opened 2020-07-01
Implement login & logout functionality

winniewan:
Implement the authentication for the webpage by using Users API to send a get request to check if user is logged in and return the correct login/logout URL. If user is a new user signing in, add user to datastore.

Since Users API log user out of all Google accounts, there is a logoutServlet that handles this situation. 
The logoutServlet will remove the cookies which leads to the user only logging out of our webpage and not all their Google account applications. 

Add Unit tests for servlets.


capstone/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,18 @@
+package com.google.sps.data;
+
+/** Individual user login data. */
+public final class User {
+  private final boolean isLoggedin;
+  private final String url;
```
prabenzom, line 6: nit: can you add a comment describing what the url string means?

capstone/src/test/java/com/google/sps/LoginServletTest.java
```diff
@@ -0,0 +1,114 @@
...
+    JsonElement responseJsonElement = new JsonParser().parse(responseString);
+
+    Gson gson = new GsonBuilder().create();
+    JsonElement userJsonElement = gson.toJsonTree(LoggedInUser);
+
+    JsonObject responseJsonObject = responseJsonElement.getAsJsonObject();
+    JsonObject userJsonObject = userJsonElement.getAsJsonObject();
+
+    Assert.assertEquals(responseJsonObject, userJsonObject);
+    helper.tearDown();
```
prabenzom, line 75: can you call helper.setup and helper.teardown in the setup and teardown functions?  I think you 
 can create an @After annotated method and call it tearDown()? 

https://cloud.google.com/appengine/docs/standard/java/tools/localunittesting/javadoc/com/google/appengine/tools/development/testing/LocalServiceTestHelper#class-localservicetesthelper

capstone/src/main/java/com/google/sps/data/User.java
winniewan: Done

capstone/src/test/java/com/google/sps/LoginServletTest.java
winniewan: Done

capstone/src/test/java/com/google/sps/servlets/authentication/NewUserServletTest.java
```diff
@@ -0,0 +1,111 @@
...
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.KeyFactory;
+import com.google.appengine.api.users.User;
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.tools.development.testing.LocalServiceTestHelper;
+import com.google.appengine.tools.development.testing.LocalUserServiceTestConfig;
+import java.io.*;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.io.StringWriter;
```
eash130, line 20: nit: you don't need to import other things from io if you just do *

capstone/src/test/java/com/google/sps/servlets/authentication/LoginServletTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+  /*
+   *  Test for when user is logged in, it should return isUserLoggedin boolean value
+   *  true and the logout URL.
+   **/
+  @Test
+  public void loggedInUserReturnsLogOutUrl() throws ServletException, IOException {
+    StringWriter stringWriter = new StringWriter();
+    PrintWriter printWriter = new PrintWriter(stringWriter);
+    when(response.getWriter()).thenReturn(printWriter);
```
eash130, line 63: nit: you could set this up in the setup method since it is used across tests.

capstone/src/test/java/com/google/sps/servlets/authentication/NewUserServletTest.java
winniewan: done

capstone/src/test/java/com/google/sps/servlets/authentication/LoginServletTest.java
winniewan: I've added that to other test classes, but not this one. 
There seems to be an error if it's done that way. The helper is also changing from one test method to another. I will leave it as is for this test class.

PR #29, opened 2020-07-07
Implement CommentServlet.java

lukasberglund:


capstone/src/main/java/com/google/sps/data/Comment.java
```diff
@@ -0,0 +1,36 @@
...
+package com.google.sps.data;
+
+public final class Comment {
+
+  private final long id;
+  private final String content;
+  private final long timestamp;
+  private final long userId;
+  private final long businessId;
+  // If comment does not have a parent a -1 is stored
```
prabenzom, line 24: i think it's fine for the default to be 0

capstone/src/test/java/com/google/sps/CommentServletTest.java
```diff
@@ -0,0 +1,179 @@
...
+
+package com.google.sps.servlets;
+
+import static com.google.appengine.api.datastore.FetchOptions.Builder.withLimit;
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.doReturn;
+
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+// import com.google.appengine.api.datastore.Entity;
```
eash130, line 23: You can remove this if you're not using.

capstone/src/test/java/com/google/sps/CommentServletTest.java
```diff
@@ -0,0 +1,179 @@
...
+      DatastoreService ds, String content, long userId, long businessId, long parentId) {
+    return ds.prepare(queryComment(content, userId, businessId, parentId))
+        .countEntities(withLimit(COUNTING_LIMIT));
+  }
+
+  @Test
+  public void testBasicDoPost() throws IOException {
+    DatastoreService ds = DatastoreServiceFactory.getDatastoreService();
+
+    setMockRequestParameters(request, MOCK_CONTENT, MOCK_USER_ID, MOCK_BUSINESS_ID, MOCK_PARENT_ID);
```
eash130, line 113: you might be able to put this in setUp since you are calling it for each test.

capstone/src/main/java/com/google/sps/servlets/CommentServlet.java
```diff
@@ -0,0 +1,76 @@
...
+      verifyRequestParameters(request);
+      storeEntity(buildCommentEntity(request));
+    } catch (IOException e) {
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST, e.toString());
+    }
+  }
+
+  /** Verify that the request contains all the necessary parameters */
+  private void verifyRequestParameters(HttpServletRequest request) throws IOException {
+    List<String> requiredParameters = new LinkedList<String>();
```
prabenzom, line 45: this seems like it could be a static final class variable on the servlet. We shouldn't instantiate this on every request.

capstone/src/test/java/com/google/sps/CommentServletTest.java
```diff
@@ -0,0 +1,179 @@
...
+      servlet.doPost(request, response);
+
+      // Check that the server rejected the request
+      Mockito.verify(response, Mockito.times(1))
+          .sendError(
+              Mockito.eq(HttpServletResponse.SC_BAD_REQUEST),
+              ArgumentMatchers.eq(
+                  "java.io.IOException: Parameter \'"
+                      + excludedParameterName
+                      + "\' missing in request."));
```
eash130, line 171: You could change this to Mockito.anyString() if you like in case the error message ever changes. We're mainly checking for the error code.

capstone/src/test/java/com/google/sps/CommentServletTest.java
```diff
@@ -0,0 +1,179 @@
...
+    DatastoreService ds = DatastoreServiceFactory.getDatastoreService();
+
+    setMockRequestParameters(request, MOCK_CONTENT, MOCK_USER_ID, MOCK_BUSINESS_ID, MOCK_PARENT_ID);
+
+    assertEquals(
+        0,
+        countCommentOccurences(ds, MOCK_CONTENT, MOCK_USER_ID, MOCK_BUSINESS_ID, MOCK_PARENT_ID));
+
+    servlet.doPost(request, response);
+    servlet.doPost(request, response);
```
eash130, line 137: Why are you testing the same comment twice?

capstone/src/main/java/com/google/sps/servlets/CommentServlet.java
```diff
@@ -0,0 +1,76 @@
...
+@WebServlet("/comment/*")
+public class CommentServlet extends HttpServlet {
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    try {
+      verifyRequestParameters(request);
+      storeEntity(buildCommentEntity(request));
+    } catch (IOException e) {
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST, e.toString());
```
prabenzom, line 39: I don't think it's a good practice to just output the exception stack trace to the user. You risk revealing too much about your implementation/may leak sensitive information. You should manually write out the error  string depending on the error.

capstone/src/main/java/com/google/sps/servlets/CommentServlet.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+
+  /** Verify that the request contains all the necessary parameters */
+  private void verifyRequestParameters(HttpServletRequest request) throws IOException {
+    List<String> requiredParameters = new LinkedList<String>();
+
+    requiredParameters.add(DatastoreNames.CONTENT_PROPERTY);
+    requiredParameters.add(DatastoreNames.USER_ID_PROPERTY);
```
prabenzom, line 48: do we need to verify that the logged in user is the user making this request?

capstone/src/main/java/com/google/sps/data/Comment.java
```diff
@@ -0,0 +1,36 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+public final class Comment {
+
+  private final long id;
```
winniewan, line 19: nit: can you place a param comment on what these variables represent?

capstone/src/test/java/com/google/sps/CommentServletTest.java
```diff
@@ -0,0 +1,179 @@
...
+                    new FilterPredicate(
+                        DatastoreNames.PARENT_ID_PROPERTY, FilterOperator.EQUAL, parentId))));
+  }
+
+  private int countCommentOccurences(
+      DatastoreService ds, String content, long userId, long businessId, long parentId) {
+    return ds.prepare(queryComment(content, userId, businessId, parentId))
+        .countEntities(withLimit(COUNTING_LIMIT));
+  }
+
```
winniewan, line 108: nit: can you add documentation to each test function? 
ex: what you're testing and what it should return

capstone/src/main/java/com/google/sps/servlets/CommentServlet.java
```diff
@@ -0,0 +1,76 @@
...
+    commentEntity.setProperty("timestamp", System.currentTimeMillis());
+    commentEntity.setProperty("userId", Long.parseLong(request.getParameter("userId")));
+    commentEntity.setProperty("businessId", Long.parseLong(request.getParameter("businessId")));
+    commentEntity.setProperty("parentId", Long.parseLong(request.getParameter("parentId")));
+
+    return commentEntity;
+  }
+
+  private void storeEntity(Entity entity) {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
```
winniewan, line 73: You can make this into a global instance and call datastore.put() directly

capstone/src/test/java/com/google/sps/CommentServletTest.java
lukasberglund: Isn't that clear based on the assertions made?

lukasberglund: Isn't that clear from reading the code? E.g. when it says
```
assertEquals(
        1,
        countCommentOccurences(ds, MOCK_CONTENT, MOCK_USER_ID, MOCK_BUSINESS_ID, MOCK_PARENT_ID));
```
It's clear that the occurences in the datastore should be one right?

winniewan: In my opinion, it's good to have documentation for classes/tests/functions. Since we're building a codebase, it would make it more organized and easier to understand without going into depth of each line.

capstone/src/test/java/com/google/sps/CommentServletTest.java
prabenzom: it is once for each test

capstone/src/test/java/com/google/sps/CommentServletTest.java
lukasberglund: I understand that, but I also don't want to write redundant comments.

capstone/src/test/java/com/google/sps/CommentServletTest.java
eash130: Maybe you could try returning after you send the error? That would prevent it from hitting the second error later on in the servlet.

capstone/src/main/java/com/google/sps/servlets/CommentServlet.java
lukasberglund: Done

capstone/src/test/java/com/google/sps/CommentServletTest.java
lukasberglund: Ah good point

lukasberglund: Nope, that didn't work either.

eash130: When you tried it with anyString(), did you put it inside eq()?

lukasberglund: I did it exactly the way you did it.

lukasberglund:
@googlebot I fixed it

lukasberglund:
@googlebot I fixed it.

PR #30, opened 2020-07-07
Utilize businesses fetched from datastore

eash130:


capstone/src/main/webapp/js/business.js
```diff
@@ -14,4 +14,20 @@
 
 import { loadCommentSection } from '/js/comments.js';
 
-window.onload = () => loadCommentSection(document.getElementById('comment-section'));
+window.onload = function() {
+  const url = new URLSearchParams(window.location.search);
```
lukasberglund, line 18: What does this do?

eash130: Could be helpful to add a comment here.

capstone/src/main/webapp/js/business.js
```diff
@@ -14,4 +14,22 @@
 
 import { loadCommentSection } from '/js/comments.js';
 
-window.onload = () => loadCommentSection(document.getElementById('comment-section'));
+window.onload = function() {
+  // window.location.search returns the search string of the URL.
+  // In this case, window.location.search = ?id={businessID}
+  const url = new URLSearchParams(window.location.search);
```
lukasberglund, line 20: Why is this `const` but `businessId` is declared with `let`?

lukasberglund:
This is great. If you could just resolve my one question that would be aweomse.

capstone/src/main/webapp/js/business.js
eash130: no reason. I might as well use two const.

PR #32, opened 2020-07-09
Refactor BusinessServlet and BusinessesServlet

winniewan:
BusinessServlet is responsible for fetching and updating specific business IDs from URL Params 

BusinessesServlet is responsible for listing all businesses profiles.


capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -30,51 +16,109 @@
...
-/** Servlet that handles business information. */
+/** Servlet responsible for showing a business user profile. */
 @WebServlet("/business/*")
 public class BusinessServlet extends HttpServlet {
 
-  private static final String TASK_NAME = "Business";
-  private static final String NAME_PROPERTY = "name";
-  private static final String EMAIL_PROPERTY = "email";
-  private static final String BIO_PROPERTY = "bio";
-  private static final String LOCATION_PROPERTY = "location";
```
eash130, line 41: I think it's fine to leave these in to be used. If any property name changes, then this is the easiest way to change it. Also, Alyssa recommended making them like this when I was making my portfolio

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -30,51 +16,109 @@
...
-      String location = (String) businessEntity.getProperty(LOCATION_PROPERTY);
-      BusinessProfile business = new BusinessProfile(id, name, email, bio, location);
-
-      Gson gson = new Gson();
-      String jsonBusiness = gson.toJson(business);
-      response.setContentType("application/json");
-      response.getWriter().println(jsonBusiness);
-    } else {
+    // Obtain userId from param URL.
+    String[] pathSegments = request.getPathInfo().split("/");
```
eash130, line 37: It might be better to just use substring(1) since you never use pathSegments[0] anyways since it is just an empty string. After that, you can check if pathSegments == "/" in which case an id was not provided.

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -30,51 +16,109 @@
...
+
+    // Create a profile object that contains the properties.
+    BusinessProfile profile =
+        new BusinessProfile(id, name, location, bio, story, about, support, isCurrentUser);
+
+    // Send it back to client side as a JSON file.
+    response.setContentType("application/json;");
+    Gson gson = new Gson();
+    response.getWriter().println(gson.toJson(profile));
+  }
```
eash130, line 89: I feel like there's a few extra steps here that was condensed more in the previous version.

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -30,51 +16,109 @@
...
+
+    // Query all profile properties.
+    String id = entity.getKey().getName();
+    String name = entity.hasProperty("name") ? (String) entity.getProperty("name") : "Anonymous";
+    String location = entity.hasProperty("location") ? (String) entity.getProperty("location") : "";
+    String bio = entity.hasProperty("bio") ? (String) entity.getProperty("bio") : "";
+    String story = entity.hasProperty("story") ? (String) entity.getProperty("story") : "";
+    String about = entity.hasProperty("about") ? (String) entity.getProperty("about") : "";
+    String support = entity.hasProperty("support") ? (String) entity.getProperty("support") : "";
+    boolean isCurrentUser = userId.equals(id);
```
eash130, line 79: Can't we assume that by getting to this point, the business has all of these properties?
If the entity is in the datastore and is a business, then it SHOULD have all of these properties at least saved in the datastore.

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -30,51 +16,109 @@
...
+    Gson gson = new Gson();
+    response.getWriter().println(gson.toJson(profile));
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String id = userService.getCurrentUser().getUserId();
+
+    // Check if user is logged in.
+    if (id == null) {
```
eash130, line 96: If they user is not logged in, then wouldn't userService.getCurrentUser() return null, which would cause a nullpointerexception on line 93?

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -30,51 +16,109 @@
...
+    }
+
+    // Mandatory property "name" needs to be filled out. If not, send an error.
+    if (request.getParameter("name") == null) {
+      response.sendError(
+          HttpServletResponse.SC_NOT_FOUND, "Required field: name was not filled out.");
+      return;
+    }
+
+    // Update properties in datastore.
```
eash130, line 109: One thing I missed on the last PR is that we should store the business email in the datastore as well so that we can display their calendar and provide a link that users can send them emails at.

capstone/src/main/java/com/google/sps/servlets/BusinessesServlet.java
```diff
@@ -17,49 +17,57 @@
...
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    // Retrieve the name and bio of all businesses to be displayed on the page.
-    Query businessQuery = new Query(TASK_NAME);
-    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    PreparedQuery queryResults = datastore.prepare(businessQuery);
-    ArrayList<BusinessProfile> businesses = new ArrayList();
+    // Query profile entities from datastore.
+    Filter propertyFilter = new FilterPredicate("isBusiness", FilterOperator.EQUAL, "Yes");
+    Query query = new Query("UserProfile").setFilter(propertyFilter);
```
eash130, line 46: You could combine these two lines.

capstone/src/main/java/com/google/sps/servlets/BusinessesServlet.java
```diff
@@ -17,49 +17,57 @@
...
-      BusinessProfile business = new BusinessProfile(id, name, email, bio, location);
-      businesses.add(business);
+    PreparedQuery results = datastore.prepare(query);
+
+    // Retreive entities with a filter on business users.
+    List<Entity> entities = results.asList(FetchOptions.Builder.withDefaults());
+
+    // Convert entities to Profile objects.
+    List<BusinessProfile> profiles = new ArrayList<>();
+    for (Entity entity : entities) {
```
eash130, line 55: you can simplify this by putting results.asIterable() here.

capstone/src/test/java/com/google/sps/servlets/BusinessServletTest.java
```diff
@@ -1,126 +1,285 @@
...
 import org.mockito.Mock;
 import org.mockito.Mockito;
 import org.mockito.MockitoAnnotations;
 
-/** Unit tests for BusinessServlet. */
+@RunWith(JUnit4.class)
 public class BusinessServletTest {
 
-  private final LocalServiceTestHelper helper =
-      new LocalServiceTestHelper(new LocalDatastoreServiceTestConfig());
```
eash130, line 43: Is the datastore service test helper not needed when dealing with datastore?

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
winniewan: I've that method before, but this way helps check the edge case for when there's additional path info after.

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
winniewan: We only have one required field, so some fields can be empty. Also, when the account is created, these fields will be empty as well.

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
winniewan: Mhm I'm not sure what you mean...

capstone/src/main/java/com/google/sps/servlets/BusinessesServlet.java
winniewan: I think that would make it too long. I think this way makes it more readable

capstone/src/test/java/com/google/sps/servlets/BusinessServletTest.java
winniewan: I think there's multiple ways to do it. Both way works I believe

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
winniewan: Great suggestion! Thanks, done

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
winniewan: I will include a try/catch for this case. Great! thanks, done.

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
winniewan: I can follow up on this in another PR :)

capstone/src/main/java/com/google/sps/servlets/BusinessesServlet.java
winniewan: Great suggestion! done.

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
eash130: When you create the business, isn't the property still there with a value of "" for any field not filled in?

winniewan: Nope, it will be null

eash130: in the POST, you set the property regardless if the fields are empty or filled. The default value for an empty field is "" (if not, then you can just set default="")

eash130: Do you think that setting the default property of the input fields in the form might resolve the problem of setting null?

winniewan: Sure! done.

prabenzom: Is there an annotation on the database layer that ensures the property is set? If so, i'd rely on that. I think it's better to be defensive in these types of situations. If the database does not ensure that that field is set then you shouldn't assume it. Even if your current code guarantees that the property is the empty string, scripts, older code, future code, etc. may not.

capstone/src/main/java/com/google/sps/servlets/BusinessesServlet.java
```diff
@@ -19,47 +19,51 @@
...
-      BusinessProfile business = new BusinessProfile(id, name, email, bio, location);
-      businesses.add(business);
+    PreparedQuery results = datastore.prepare(query);
+
+    // Convert entities to Profile objects.
+    List<BusinessProfile> profiles = new ArrayList<>();
+    for (Entity entity : results.asIterable()) {
+      String id = (String) entity.getKey().getName();
+      String name = (String) entity.getProperty("name");
+      String location = (String) entity.getProperty("location");
```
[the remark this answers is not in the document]
winniewan: I was planning to set default values (as "") for these properties when it's a new user. When the user edits their profile and they decide to leave certain properties blank, it will be caught by the doPost and default it to "". So datastore will always have a default value for these properties.

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -30,51 +16,134 @@
...
+    Gson gson = new Gson();
+    response.getWriter().println(gson.toJson(profile));
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String id;
+
+    // Check if user is logged in.
+    try {
```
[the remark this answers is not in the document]
winniewan: SG, done

capstone/src/main/java/com/google/sps/servlets/BusinessesServlet.java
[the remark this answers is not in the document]
prabenzom: I was referring to "NAME_PROPERTY" and "EMAIL_PROPERTY". Why don't we use those to get the properties from the entity?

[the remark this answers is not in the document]
winniewan: Oh I mistaken businessesServlet with businessServlet.

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -30,51 +16,136 @@
...
+        new BusinessProfile(id, name, location, bio, story, about, support, isCurrentUser);
+
+    // Send it back to client side as a JSON file.
+    response.setContentType("application/json;");
+    Gson gson = new Gson();
+    response.getWriter().println(gson.toJson(profile));
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
eash130, line 102: I noticed this doPost is never being used. Just leave it in the PR, I'll add this to my code and then make it get used.

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -30,51 +16,136 @@
...
+          "The profile you were looking for was not found in our records!");
+      return;
+    }
+
+    // If userId is not a business owner id, redirect to "profile not found" page.
+    String isBusiness =
+        entity.hasProperty(IS_BUSINESS_PROPERTY)
+            ? (String) entity.getProperty(IS_BUSINESS_PROPERTY)
+            : "";
+    if (isBusiness.equals("No")) {
```
alygerharter, line 67: if isBusiness is empty ("") should that also return NOT_FOUND?

capstone/src/test/java/com/google/sps/servlets/BusinessServletTest.java
```diff
@@ -1,126 +1,271 @@
...
+    Assert.assertEquals(capEntity.getProperty("name"), NAME);
+    Assert.assertEquals(capEntity.getProperty("location"), LOCATION);
+    Assert.assertEquals(capEntity.getProperty("bio"), BIO);
+    Assert.assertEquals(capEntity.getProperty("story"), STORY);
+    Assert.assertEquals(capEntity.getProperty("about"), ABOUT);
+    Assert.assertEquals(capEntity.getProperty("support"), SUPPORT);
+  }
+
+  // Create an entity with USERID.
+  public Entity createEntity() {
```
prabenzom, line 253: can these be more descriptive? createuserProfile

capstone/src/test/java/com/google/sps/servlets/BusinessServletTest.java
```diff
@@ -1,126 +1,271 @@
...
+    String keyString = KeyFactory.createKeyString("UserProfile", USER_ID);
+    Key userKey = KeyFactory.stringToKey(keyString);
+    return new Entity("UserProfile", USER_ID);
+  }
+
+  // Set static variables to entity.
+  public Entity setEntity() {
```
prabenzom, line 260: maybe set userProfileData

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -30,51 +16,136 @@
...
+    if (request.getParameter(NAME_PROPERTY) == null) {
+      response.sendError(
+          HttpServletResponse.SC_NOT_FOUND, "Required field: name was not filled out.");
+      return;
+    }
+
+    // Update properties in datastore.
+    Entity businessEntity = new Entity("UserProfile", id);
+
+    businessEntity.setProperty(IS_BUSINESS_PROPERTY, getParam(IS_BUSINESS_PROPERTY, request));
```
prabenzom, line 122: Is this something clients should be able to set? This means the front end could use the business servlet to create a user profile right?

prabenzom: also, add a test what happens if the client sets isBusiness = no

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -30,51 +16,136 @@
...
+          HttpServletResponse.SC_NOT_FOUND, "You don't have permission to perform this action!");
+      return;
+    }
+
+    String id = userService.getCurrentUser().getUserId();
+
+    // Mandatory property "name" needs to be filled out. If not, send an error.
+    if (request.getParameter(NAME_PROPERTY) == null) {
+      response.sendError(
+          HttpServletResponse.SC_NOT_FOUND, "Required field: name was not filled out.");
```
eash130, line 115: I think SC_BAD_REQUEST is more appropriate that NOT_FOUND here

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
winniewan: SG, done.

capstone/src/test/java/com/google/sps/servlets/BusinessServletTest.java
winniewan: yep, thanks! done.

capstone/src/test/java/com/google/sps/servlets/BusinessServletTest.java
winniewan: done

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
winniewan: That's a great catch!

PR #33, opened 2020-07-10
Implement Commentsservlet.java

lukasberglund:


lukasberglund:
I'm sorry that some changes to other files like the javascript and `CommentServletTest.java` snuck their way in. I'll fix the javascript in an upcoming PR.

lukasberglund:
@googlebot I fixed it.

capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
```diff
@@ -0,0 +1,146 @@
...
+  /**
+   * Converts a Comment instance into a JSON string using the Gson library. Note: We first added the
+   * Gson library dependency to pom.xml.
+   */
+  private String convertToJson(List<Comment> comments) {
+    Gson gson = new Gson();
+    String json = gson.toJson(comments);
+
+    return json;
+  }
```
eash130, line 145: I don't think this helper is needed since you only convertToJson once in this file and it's not that much code.

capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
```diff
@@ -0,0 +1,146 @@
...
+
+  private List<Entity> runCommentsQuery(HttpServletRequest request)
+      throws IllegalArgumentException {
+    Query query =
+        new Query(DatastoreNames.COMMENT_ENTITY_NAME)
+            .setFilter(buildFilter(request))
+            .addSort(DatastoreNames.TIMESTAMP_PROPERTY, SortDirection.DESCENDING);
+
+    return datastore.prepare(query).asList(FetchOptions.Builder.withLimit(COMMENT_LIMIT));
+  }
```
eash130, line 101: Might not need a helper here since it's only called once and it's not too complex to just have in doGet.

capstone/src/main/webapp/js/comments.js
```diff
@@ -12,11 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { wrapInPromise, buildElement, buildButton } from '/js/util.js';
+import { wrapInPromise, buildElement, buildButton, postRequest } from '/js/util.js';
 
 var commentContainer = undefined;
 var commentField = undefined;
 
+const NULL_VAL = -1;
```
eash130, line 20: I might have missed the resolution, but I think Ben said that using 0 for a null ID works.

capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
lukasberglund: I think having a helper makes it more readable.

capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
lukasberglund: Fair. I'll remove it.

capstone/src/main/webapp/js/comments.js
lukasberglund: I went ahead and reverted all the changes I had made to javascript.

capstone/src/test/java/com/google/sps/servlets/CommentsServletTest.java
```diff
@@ -0,0 +1,287 @@
...
+  private void assertResponseWithArbitraryTextRaised(int targetResponse) throws IOException {
+    Mockito.verify(response, Mockito.times(1))
+        .sendError(Mockito.eq(targetResponse), Mockito.anyString());
+  }
+
+  /**
+   * Runs a standard test in which one parameter in the request is specified, the datastore has the
+   * * same data in it that it always does, and a certain json response is expected.
+   */
+  private void runTest(
```
prabenzom, line 148: can you use parameterized tests to accomplish this? https://blog.parasoft.com/how-to-create-junit-parameterized-tests-faster

capstone/src/test/java/com/google/sps/servlets/CommentsServletTest.java
```diff
@@ -0,0 +1,287 @@
...
+    doReturn(new PrintWriter(servletResponseWriter)).when(response).getWriter();
+    servlet = new CommentsServlet(ds);
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  private void setParameter(String parameterName, String parameterVal) {
```
prabenzom, line 71: i don't know if this is worth having a helper method

capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
prabenzom: I think this is fine to have in a helper method 

Per the guidelines: 

This is not an exact science, but here are some examples of signs that your method might be too complex:

It has several levels of nesting 
It consists of several separate blocks of code, that each require a comment to explain what it does.
It has to handle different types of exceptions in different places.
It can't all fit on your screen at once.
The remedy is to extract one or more smaller, more focused, clearly-named private helper methods. Note that there is nothing wrong with a private helper method that is used only once; these can be a useful tool for making code more readable.

capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
```diff
@@ -0,0 +1,135 @@
...
+      String jsonComments = new Gson().toJson(comments);
+
+      response.setContentType("application/json;");
+      response.getWriter().println(jsonComments);
+    } catch (IllegalArgumentException e) {
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST, INVALID_ARGUMENT_MESSAGE);
+    }
+  }
+
+  private List<Entity> runCommentsQuery(HttpServletRequest request)
```
prabenzom, line 93: only pass in the parameters that are necessary for this method. it looks like this is just property name and filter parameter. you don't need anything else from the request.

capstone/src/test/java/com/google/sps/servlets/CommentsServletTest.java
lukasberglund: Don't you think it improves ease of reading?

capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
lukasberglund: How would you do this? Would you just pass the filter in directly here?

capstone/src/main/java/com/google/sps/servlets/CommentServlet.java
```diff
@@ -68,6 +68,8 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     // Verify that the user posting the comment is the user is who's logged in
     User currentUser = userService.getCurrentUser();
     if (currentUser == null || !currentUser.getUserId().equals(request.getParameter("userId"))) {
+      System.out.println("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!");
```
winniewan, line 71: Haha is this a debug print statement?

lukasberglund: Yeah I was showing Ben something lol

capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
```diff
@@ -0,0 +1,135 @@
...
+      throws IllegalArgumentException, IOException {
+    try {
+      List<Entity> entities = runCommentsQuery(request);
+      List<Comment> comments =
+          entities.stream().map(entity -> generateComment(entity)).collect(Collectors.toList());
+
+      String jsonComments = new Gson().toJson(comments);
+
+      response.setContentType("application/json;");
+      response.getWriter().println(jsonComments);
```
winniewan, line 87: You can take these out of the try-catch

lukasberglund: Tru

capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
prabenzom: runCommentsQuery(string propertyName, string filterParameter).

capstone/src/test/java/com/google/sps/servlets/CommentsServletTest.java
prabenzom: I personally don't. one line of setParameter(name,val) is about the same as one line of doReturn(parameterVal).when(request).getParameter(parameterName); imo.

The latter makes it more obvious what we are doing (setting a mock) so i'd actually prefer that.

lukasberglund: Okay I'll change it.

capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
lukasberglund: I changed it. Lmk if this is what you had in mind :)

PR #34, opened 2020-07-13
Remodel Profile Servlet and Unit Tests

winniewan:


capstone/src/main/java/com/google/sps/servlets/profile/ProfileServlet.java
```diff
@@ -19,18 +19,15 @@
 /** Servlet responsible for showing a non-business user profile. */
 @WebServlet("/profile/*")
 public class ProfileServlet extends HttpServlet {
+  private static final String IS_BUSINESS_PROPERTY = "isBusiness";
+  private static final String NAME_PROPERTY = "name";
+  private static final String LOCATION_PROPERTY = "location";
+  private static final String BIO_PROPERTY = "bio";
```
lukasberglund, line 25: You might consider putting this in a file called `ProfileDatastore.java` so that you don't have to write these declarations in every file you where use these names.

capstone/src/main/java/com/google/sps/servlets/profile/ProfileServlet.java
```diff
@@ -86,32 +89,48 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    return request.getParameter(property);
+  }
+
+  public String getNameParam(String name, HttpServletRequest request) {
+    if (request.getParameter(name) == null) {
+      return "Anonymous";
+    }
+
+    return request.getParameter(name);
+  }
```
lukasberglund, line 135:
```suggestion
  public String getNameParam(HttpServletRequest request) {
    if (request.getParameter(NAME_PROPERTY) == null) {
      return "Anonymous";
    }

    return request.getParameter(name);
  }
```
I think this would be better since this function is specifically designed to get the name property.

capstone/src/test/java/com/google/sps/servlets/profile/ProfileServletTest.java
```diff
@@ -1,25 +1,24 @@
 package com.google.sps.servlets.profile;
 
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
lukasberglund, line 3: The java styleguide [doesn't allow wildcard](https://google.github.io/styleguide/javaguide.html#s3.3-import-statements) imports so you should replace this one with the specific imports you need.

capstone/src/test/java/com/google/sps/servlets/profile/ProfileServletTest.java
```diff
@@ -1,25 +1,24 @@
...
 import com.google.appengine.tools.development.testing.LocalServiceTestHelper;
 import com.google.appengine.tools.development.testing.LocalUserServiceTestConfig;
+import com.google.common.collect.ImmutableMap;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
 import com.google.sps.data.UserProfile;
 import java.io.*;
```
lukasberglund, line 20: Same here

capstone/src/test/java/com/google/sps/servlets/profile/ProfileServletTest.java
```diff
@@ -60,10 +45,31 @@
...
+    helper =
+        new LocalServiceTestHelper(
+                new LocalUserServiceTestConfig(), new LocalDatastoreServiceTestConfig())
+            .setEnvEmail(EMAIL)
+            .setEnvAuthDomain(AUTHDOMAIN)
+            .setEnvIsLoggedIn(true)
+            .setEnvAttributes(
+                new HashMap(
+                    ImmutableMap.of(
+                        "com.google.appengine.api.users.UserService.user_id_key", USER_ID)));
```
lukasberglund, line 68: I'm just curious, why did you choose to initialize it here rather than during declaration of the `helper` variable?

winniewan: There's some test cases that test out if a user is invalid or not found.

capstone/src/main/java/com/google/sps/servlets/profile/ProfileServlet.java
```diff
@@ -86,32 +89,48 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+
+  public String getParam(String property, HttpServletRequest request) {
+    if (request.getParameter(property) == null) {
+      return "";
+    }
+
+    return request.getParameter(property);
+  }
+
+  public String getNameParam(String name, HttpServletRequest request) {
```
prabenzom, line 129: Why do we need this function? Isn't name a required field? it should always be set

capstone/src/main/java/com/google/sps/servlets/profile/ProfileServlet.java
winniewan: I can follow up on this later :) thanks!

capstone/src/main/java/com/google/sps/servlets/profile/ProfileServlet.java
winniewan: Good catch! Thanks, done.

capstone/src/test/java/com/google/sps/servlets/profile/ProfileServletTest.java
winniewan: Thanks! done

capstone/src/test/java/com/google/sps/servlets/profile/ProfileServletTest.java
winniewan: done

capstone/src/main/java/com/google/sps/servlets/profile/ProfileServlet.java
winniewan: Whoops, right! I checked that case already. Removed, done.

PR #35, opened 2020-07-13
Structure Business related servlets

eash130:
Some things here are from Winnie's PR, such as doPost in BusinessServlet.


capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -34,38 +37,47 @@
...
-  private static final String EMAIL_PROPERTY = "email";
   private static final String BIO_PROPERTY = "bio";
   private static final String LOCATION_PROPERTY = "location";
+  private static final String STORY_PROPERTY = "story";
+  private static final String ABOUT_PROPERTY = "about";
+  private static final String SUPPORT_PROPERTY = "support";
+  private static final String CALENDAR_PROPERTY = "calendarEmail";
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
eash130, line 51: This is how I would implement doGet for BusinessServlet with the new items. Feel free to compare to Winnie's PR and leave feedback

eash130: It uses query filters to retrieve the appropriate item from datastore and also condenses a few of the conditional checks.

capstone/src/main/webapp/js/profile.js
```diff
@@ -41,22 +41,22 @@ window.toggleProfile = function() {
...
 
   if (isBusiness.checked == true) {
-    basicQuesionnaire.style.display = 'block';
     businessQuesionnaire.style.display = 'block';
+    profileForm.action = '/business';
   } 
   if (isNotBusiness.checked == true) {
-    basicQuesionnaire.style.display = 'block';
     businessQuesionnaire.style.display = 'none';
+    profileForm.action = '/profile';
```
eash130, line 56: These are the main changes apart from changing " -> ' and snake case -> camel case

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -77,4 +89,55 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
           "The business you were looking was not found in our records!");
     }
   }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
eash130, line 94: This method is from Winnie's PR apart from using Objects.toString for null values and the addition of the calendar email property

capstone/src/main/java/com/google/sps/data/BusinessProfile.java
```diff
@@ -2,17 +2,31 @@
...
+      String calendarEmail,
+      String bio,
+      String location,
+      String story,
+      String about,
+      String support) {
     this.id = id;
     this.name = name;
-    this.email = email;
+    this.calendarEmail = calendarEmail;
```
lukasberglund, line 25: Why change the name to `calendarEmail`? What if we end up using the email for other purposes?

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -77,4 +89,55 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      response.sendError(
+          HttpServletResponse.SC_BAD_REQUEST, "Required field: name was not filled out.");
+      return;
+    }
+
+    // Update properties in datastore.
+    Entity businessEntity = new Entity(USER_TASK, id);
+
+    businessEntity.setProperty(
+        IS_BUSINESS_PROPERTY, Objects.toString(request.getParameter(IS_BUSINESS_PROPERTY), ""));
```
lukasberglund, line 116: You might consider putting this into a helper function. You would save 5 lines of code and IMO it would be either to read. On the other hand, it seems like I use helper functions more than Ben would want me to so I might be acting too eagerly here.

capstone/src/main/java/com/google/sps/data/BusinessProfile.java
eash130: That's a fair point. I think it could be helpful to just keep it in. I changed it to calendarEmail because I didn't really see it being used anywhere else so I just made it more specific to my use case.

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
eash130: Looking at it, I think a helper could make sense here

capstone/src/main/java/com/google/sps/data/BusinessProfile.java
prabenzom: Per our discussion, this is also not necessarily the user's email address right? Can you add a comment indicating what this email is exactly?

alygerharter: I think just `email` is fine here, unless you anticipate a scenario where the email for the account and the email for the calendar could be different.

capstone/src/main/java/com/google/sps/servlets/profile/ProfileServlet.java
```diff
@@ -59,8 +60,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     }
 
     // If userId is a business owner id, redirect to "profile not found" page.
-    String isBusiness =
-        entity.hasProperty("isBusiness") ? (String) entity.getProperty("isBusiness") : "";
+    String isBusiness = Objects.toString(entity.getProperty("isBusiness"), "");
```
alygerharter, line 63: Can you statically import Objects.toString?

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
prabenzom: Not opposed to it, we usually have a separate class that is responsible for reading/writing to datastore to help with readability so a helper method is fine.

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -77,4 +89,55 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    businessEntity.setProperty(
+        STORY_PROPERTY, Objects.toString(request.getParameter(STORY_PROPERTY), ""));
+    businessEntity.setProperty(
+        ABOUT_PROPERTY, Objects.toString(request.getParameter(ABOUT_PROPERTY), ""));
+    businessEntity.setProperty(
+        SUPPORT_PROPERTY, Objects.toString(request.getParameter(SUPPORT_PROPERTY), ""));
+
+    // Verify that the email can be used to generate a google calendar.
+    String calendarEmail = Objects.toString(request.getParameter(CALENDAR_PROPERTY), "");
+    if (calendarEmail.contains("@gmail.com") || calendarEmail.contains("@google.com")) {
```
prabenzom, line 131: This is fine for now, not sure how this would work for businesses that use google calendar under their own domain name just as an FYI.

eash130: The worst that would happen if I remove this check is that the calendar won't load and I'm not sure how I would go about making it general. maybe there's a way to check for google accounts from email addresses.

capstone/src/main/java/com/google/sps/data/BusinessProfile.java
eash130: It's a bit unclear here since I haven't attached the corresponding HTML with this, but calendarEmail could be distinct from email in the case that the business has another email address that is designated for events.
For example, the owner of a business registers the account on behalf of the business. The owner's calendar might not necessarily contain just the events, so the owner might opt to use another google calendar that has just the events.

capstone/src/main/java/com/google/sps/servlets/profile/ProfileServlet.java
eash130: Looking online, it doesn't seem that I can since Objects.toString won't correctly override Object.toString

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -34,38 +37,47 @@
...
+                    FilterOperator.EQUAL.of(IS_BUSINESS_PROPERTY, "Yes"),
+                    FilterOperator.EQUAL.of(
+                        Entity.KEY_RESERVED_PROPERTY,
+                        KeyFactory.createKey(USER_TASK, businessID))));
 
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
     PreparedQuery queryResults = datastore.prepare(businessQuery);
     Entity businessEntity = queryResults.asSingleEntity();
 
     if (businessEntity != null) {
```
winniewan, line 70: You can do businessEntity == null { return error }

capstone/src/main/java/com/google/sps/servlets/profile/ProfileServlet.java
alygerharter: Got it, that makes sense. Thanks for checking!

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
eash130: It doesn't seem to make a big difference since the else does handle it without having to do a return from the function

winniewan: I think it's easier to understand than a big if-statement

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
eash130: Do you think it might be more helpful if I have a general purpose helper method that handles both get and set?

capstone/src/main/java/com/google/sps/data/BusinessProfile.java
prabenzom: It can be different. Businesses might have a public calendar that is separate from the cal associated with their email address, in which case calendar provides a separate email address to use for embedding the cal

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
prabenzom: For readability, the preference is to have smaller if statements.. So  I think if businessEntity == null {return error} is better.

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -20,131 +22,111 @@
...
-
-    try {
-      entity = datastore.get(userKey);
-    } catch (EntityNotFoundException e) {
-      System.err.println("Could not find key: " + userKey);
-      response.sendError(
-          HttpServletResponse.SC_NOT_FOUND,
-          "The profile you were looking for was not found in our records!");
-      return;
+  private void setEntityProperties(
```
alygerharter, line 39: nit - can you move this helper below doPost?

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -20,131 +22,111 @@
...
 
+    String id = userService.getCurrentUser().getUserId();
+
     // Update properties in datastore.
-    Entity businessEntity = new Entity("UserProfile", id);
+    Entity businessEntity = new Entity(USER_TASK, id);
 
     // If user is a non-business owner, return error.
-    if (getParam(IS_BUSINESS_PROPERTY, request).equals("No")) {
+    if (Objects.toString(request.getParameter(IS_BUSINESS_PROPERTY), "").equals("No")) {
```
alygerharter, line 109: what if Objects.toString returns ""? it might be better in this case to check for the positive case rather than the negative

capstone/src/test/java/com/google/sps/servlets/BusinessServletTest.java
```diff
@@ -76,220 +76,177 @@ public void setUp() throws Exception {
...
   @Test
   public void userNotInDatastoreReturnError()
       throws ServletException, IOException, EntityNotFoundException {
     when(request.getPathInfo()).thenReturn(PATHINFO);
 
-    String keyString = KeyFactory.createKeyString("UserProfile", USER_ID);
-    Key userKey = KeyFactory.stringToKey(keyString);
-    Entity ent = new Entity("UserProfile", USER_ID);
+    // Populate the datastore with a business with the wrong target ID.
+    Entity someBusiness = new Entity(USER_TASK, USER_ID + "1");
```
winniewan, line 125: You can set a static variable here instead of + "1" like USER2_ID or something similar

capstone/src/main/java/com/google/sps/servlets/BusinessServlet.java
```diff
@@ -20,131 +22,111 @@
...
+    Entity businessEntity = datastore.prepare(businessQuery).asSingleEntity();
+
+    if (businessEntity == null) {
       response.sendError(
           HttpServletResponse.SC_NOT_FOUND,
-          "The profile you were looking for was not found in our records!");
+          "The business you were looking was not found in our records!");
       return;
     }
+    String id = (String) businessEntity.getKey().getName();
```
winniewan, line 63: Can you change this to "userService.getCurrentUser().getUserId()"? 
Thanks!

capstone/src/main/java/com/google/sps/servlets/business/BusinessServlet.java
```diff
@@ -44,123 +44,104 @@
...
+  private static final String CALENDAR_PROPERTY = "calendarEmail";
+
   UserService userService = UserServiceFactory.getUserService();
 
   DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    // Obtain userId from param URL.
-    String[] pathSegments = request.getPathInfo().split("/");
```
prabenzom, line 54: why was this reverted?

capstone/src/main/java/com/google/sps/servlets/business/BusinessServlet.java
```diff
@@ -44,123 +44,104 @@
...
+    boolean isCurrentUser = businessID.equals(id);
+    BusinessProfile business =
+        new BusinessProfile(id, name, location, bio, story, about, email, support, isCurrentUser);
 
-    // Query all profile properties.
-    String id = entity.getKey().getName();
-    String name =
-        entity.hasProperty(NAME_PROPERTY)
-            ? (String) entity.getProperty(NAME_PROPERTY)
-            : "Anonymous";
```
prabenzom, line 95: why was this reverted?

capstone/src/main/java/com/google/sps/servlets/business/BusinessServlet.java
```diff
@@ -44,123 +44,104 @@
...
       return;
     }
 
+    String id = userService.getCurrentUser().getUserId();
+
     // Update properties in datastore.
     Entity businessEntity = new Entity(PROFILE_TASK_NAME, id);
 
     // If user is a non-business owner, return error.
-    if (getParam(IS_BUSINESS_PROPERTY, request).equals(NO)) {
```
prabenzom, line 139: why was this reverted? the preference is to return the error to eliminate if else.

so if business == no {
 return error
}

// rest of code here

capstone/src/main/java/com/google/sps/servlets/business/BusinessServlet.java
eash130: .split("/") -> .substring(1): This revert happened since I didn't think it was clean to use .split() since it creates an unnecessary array that consists of ["", "{id}"] everytime. I believe that this was designed based on request.getPathInfo() returning "business/{id}" when it actually returns "/{id}"

create key then get -> query for entity: This revert happened since it condenses a few of the conditionals that are being checked into one.

capstone/src/main/java/com/google/sps/servlets/business/BusinessServlet.java
eash130: By the design of the doPost(), the stored properties will either contain the value we want or an empty string. We also don't need to necessarily check to see if the properties exist since only business entities will be used at this point, which have some value assigned for each of these properties.

capstone/src/main/java/com/google/sps/servlets/business/BusinessServlet.java
eash130: Alyssa pointed out the possibility of isBusiness == "", rather than checking for both NO and "", we decided it would be easier to check for YES and send an error in the else block.

prabenzom: In that case you can just do ! equals yes then have the return block come first.

capstone/src/main/java/com/google/sps/servlets/business/BusinessServlet.java
prabenzom: Got it. And to confirm, calling from the servlet returns /{id} and not /business when you deploy this?

eash130: Yes, the full URL path is '/business/{id}' and request.getPathInfo() returns '/{id}'

capstone/src/main/java/com/google/sps/servlets/business/BusinessServlet.java
eash130: That makes more sense and keeps the if statement small as well.

capstone/src/main/java/com/google/sps/servlets/business/BusinessServlet.java
```diff
@@ -44,123 +44,104 @@
...
+    response.getWriter().println(gson.toJson(business));
   }
 
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     // Check if user is logged in.
     if (userService.getCurrentUser() == null) {
       response.sendError(
-          HttpServletResponse.SC_NOT_FOUND, "You don't have permission to perform this action!");
+          HttpServletResponse.SC_FORBIDDEN, "You don't have permission to perform this action!");
```
eash130, line 100: Would it be better to use the status code SC_UNAUTHORIZED here?

winniewan: Yeah. I think SC_UNAUTHORIZED makes sense

PR #37, opened 2020-07-14
Set up automatic cloud build scripts

lukasberglund:


alygerharter:
was this PR intended to include .java and .js files? or should it just be the .yaml?

lukasberglund:
@prabenzom @alygerharter the javascript was added by accident. I changed it back. My b.

lukasberglund:
The stuff in the java files is in googles autoformatter. For some weird reason it must not have autoformatted the previous commit.

PR #43, opened 2020-07-15
Stop requiring a userId supplied from the client side to post comments

lukasberglund:
Instead of requiring the userId to be supplied from the client side the server now independently determines the user.

This change makes it easier for comments to be posted from the client side.


capstone/src/test/java/com/google/sps/servlets/comments/CommentServletTest.java
```diff
@@ -185,12 +178,17 @@ public void testInvalidRequests() throws IOException {
...
   }
 
-  // Make sure it's impossible to post comment under a different name
   @Test
-  public void testWrongUserLoggedIn() throws IOException {
-    doReturn(MOCK_USER_ID + 1).when(request).getParameter(USER_ID_PROPERTY);
-    servlet.doPost(request, response);
+  public void testUserNotLoggedIn() throws IOException {
+    // Simulate no user being logged in
+    helper =
```
eash130, line 184: you could just do helper.setEnvIsLoggedIn(false) here instead of redefining helper

capstone/src/test/java/com/google/sps/servlets/comments/CommentServletTest.java
```diff
@@ -185,12 +178,17 @@ public void testInvalidRequests() throws IOException {
...
+        new LocalServiceTestHelper(new LocalUserServiceTestConfig())
+            .setEnvEmail(MOCK_EMAIL)
+            .setEnvAuthDomain(MOCK_DOMAIN)
+            .setEnvIsLoggedIn(false);
+    helper.setUp();
```
eash130, line 189: You can just do helper.setEnvIsLoggedIn(false); here. This would allow you to use the same helper and not have to call helper.setUp() again.

lukasberglund: Done

capstone/src/test/java/com/google/sps/servlets/comments/CommentServletTest.java
```diff
@@ -185,12 +178,13 @@ public void testInvalidRequests() throws IOException {
     }
   }
 
-  // Make sure it's impossible to post comment under a different name
   @Test
-  public void testWrongUserLoggedIn() throws IOException {
-    doReturn(MOCK_USER_ID + 1).when(request).getParameter(USER_ID_PROPERTY);
-    servlet.doPost(request, response);
+  public void testUserNotLoggedIn() throws IOException {
```
alygerharter, line 182: can you add a test that verifies the user id is set on the comment correctly?

lukasberglund: I added a test, lmk if that's what you had in mind :)

PR #49, opened 2020-07-16
Add a location auto complete option in edit profile using Maps API

winniewan:
- User can insert their location with Google Maps API auto complete help. This can be used to display their location on a Google map that will be implemented in the future. 
- Allow app engine to display server errors accordingly.


capstone/src/main/webapp/js/profile.js
```diff
@@ -60,21 +57,23 @@ window.hasAnswerQuestionnaire = function() {
 function displayProfile() {
   let id = getId();
   fetch('/profile/'+id)
-    .then(response => response.json())
+    .then(response => {
+      if (!response.ok) {
```
lukasberglund, line 61: I don't understand where this field comes from. As far as I can tell `Profile.java` does not have the field `ok`. Can you explain?

capstone/src/main/webapp/js/profile.js
```diff
@@ -60,21 +57,23 @@ window.hasAnswerQuestionnaire = function() {
...
-function displayEditProfileValues() {
-  let id = getId();
-  fetch('/profile/'+id)
-    .then(response => response.json())
-    .then((userProfile) => {
-      setEditValues(userProfile.name, userProfile.location, userProfile.bio);
-    });
+function writeErrorPage(id) {
+  location.assign('/profile/' + id);
 }
```
lukasberglund, line 77: What does this do?

capstone/src/main/webapp/js/profile.js
```diff
@@ -129,3 +128,24 @@ function setEditValues(name, location, bio) {
   location_section.value = location;
   bio_section.value = bio;
 }
+
+// Bias the autocomplete object to the user's geographical location,
+// as supplied by the browser's 'navigator.geolocation' object.
+window.geolocate = function() {
```
lukasberglund, line 134: Why not `function geolocate() {`?

winniewan: Since you're calling the function from HTML, it can be called from the window object

capstone/src/main/webapp/js/profile.js
winniewan: When the server sends a 404 error response, I'm going to redirect them to that default page

capstone/src/main/webapp/js/profile.js
winniewan: You can read more about it here: https://developer.mozilla.org/en-US/docs/Web/API/Response/ok
it gives a boolean value whether the response was successful

capstone/src/main/webapp/js/profile.js
lukasberglund: I feel like the function could have a more descriptive name. What about calling it `redirectToDefaultPage()`

capstone/src/main/webapp/js/profile.js
lukasberglund: Ahh

capstone/src/main/webapp/js/profile.js
winniewan: Sure!

PR #50, opened 2020-07-17
Add mini map to profile page

winniewan:


capstone/src/main/webapp/js/profile.js
```diff
@@ -149,3 +151,31 @@ window.geolocate = function() {
     });
   }
 }
+
+// Create the mini map on profile page.
+function createProfileMap(address) {
+  let map = new google.maps.Map(document.getElementById('map'), {
+    zoom: 8,
+    center: {lat: -34.397, lng: 150.644}
```
eash130, line 159: Might be helpful to leave a comment indicating where this center is located

winniewan: Yep! I'll add it, thanks. done

PR #53, opened 2020-07-20
Add an autocomplete option to enter your location and a mini Google map to business page.

winniewan:


capstone/src/main/webapp/business.html
```diff
@@ -99,6 +106,8 @@ <h3>How You Can Help Us</h3>
         </form> 
       </div>
 
+      <div id="map"></div>
```
eash130, line 109: I think the map might be more fittingly placed if it was below the location header.

winniewan: I think it's okay to leave it there for now. The design might be different after adding bootstrap to it

PR #58, opened 2020-07-21
Allow for users not logged in to view business pages

eash130:
I accidentally pushed this change with the typo CL, it's an extra step so that if a user that isn't logged into the website tries to navigate to a business page, they won't be encountered with a null pointer exception.

If it looks good, then I'll just close the PR.

### DIFF
--- a/capstone/src/main/java/com/google/sps/servlets/business/BusinessServlet.java
+++ b/capstone/src/main/java/com/google/sps/servlets/business/BusinessServlet.java
@@ -75,7 +75,10 @@ public class BusinessServlet extends HttpServlet {
           "The business you were looking was not found in our records!");
       return;
     }
-    String id = userService.getCurrentUser().getUserId();
+    String id = "";
+    if (userService.getCurrentUser() != null) {
+      id = userService.getCurrentUser().getUserId();
+    }
     String name = (String) businessEntity.getProperty(NAME_PROPERTY);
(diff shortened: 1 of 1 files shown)


capstone/src/main/java/com/google/sps/servlets/business/BusinessServlet.java
```diff
@@ -75,10 +75,12 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
           "The business you were looking was not found in our records!");
       return;
     }
+    // SEE BELOW
     String id = "";
     if (userService.getCurrentUser() != null) {
       id = userService.getCurrentUser().getUserId();
     }
+    // SEE ABOVE
```
lukasberglund, line 83: These comments confuse me.

eash130: the edit that i made before was on lines between the comments. I accidentally pushed this with another PR.

eash130: If the code in between looks fine, then I'll just close this PR

eash130: Looking at it now, I'm not sure why. I only added the conditional.

PR #59, opened 2020-07-21
Add map servlet and unit test for the home page

winniewan:


capstone/src/main/java/com/google/sps/servlets/MapServlet.java
```diff
@@ -0,0 +1,126 @@
...
+
+    float SW_Lat, SW_Lng, NE_Lat, NE_Lng;
+
+    try {
+      SW_Lat = coordinates.get(0);
+      SW_Lng = coordinates.get(1);
+      NE_Lat = coordinates.get(2);
+      NE_Lng = coordinates.get(3);
+    } catch (IndexOutOfBoundsException e) {
+      response.sendError(HttpServletResponse.SC_NOT_FOUND, "The map is not found.");
```
eash130, line 73: Also could use SC_BAD_REQUEST since not all coordinates were provided.

capstone/src/test/java/com/google/sps/servlets/MapServletTest.java
```diff
@@ -0,0 +1,203 @@
...
+  private static final String NAME = "Pizzeria";
+  private static final String LOCATION_IN_BOUNDS = "Mountain View, CA, USA";
+  private static final String LOCATION_NOT_IN_BOUNDS = "New York, NY";
+  private static final String BIO = "This is my business bio.";
+  private static final String STORY = "The pandemic has affected my business in X many ways.";
+  private static final String ABOUT = "Here is the Pizzeria's menu.";
+  private static final String EMAIL = "email@business.biz";
+  private static final String SUPPORT = "Please donate at X website.";
+  private static final String PATHINFO = "map/37.2227223/-122.3033039/37.548271/-121.988571";
+  private static final String INVALID_PATHINFO = "map";
```
eash130, line 76: Are you using all of these?

capstone/src/main/java/com/google/sps/servlets/MapServlet.java
```diff
@@ -0,0 +1,126 @@
...
+    Rectangle bounds = new Rectangle(SW, NE);
+
+    Filter propertyFilter =
+        new CompositeFilter(
+            CompositeFilterOperator.AND,
+            Arrays.asList(
+                new FilterPredicate(IS_BUSINESS_PROPERTY, FilterOperator.EQUAL, YES),
+                new StContainsFilter("geoPt", bounds)));
+
+    Query query = new Query(PROFILE_TASK_NAME).setFilter(propertyFilter);
```
eash130, line 89: So will the user have a geoPt propery?

capstone/src/main/java/com/google/sps/servlets/MapServlet.java
```diff
@@ -0,0 +1,126 @@
...
+@WebServlet("/map/*")
+public class MapServlet extends HttpServlet {
+
+  DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+    // Get bounds SW and NE lat and longs and replace it underneath.
+    String pathInfo = request.getPathInfo(); // {value}/SW_Lat/SW_Lng/NE_Lat/NE_Lng
```
prabenzom, line 62: I think these should be parameters of the request, not part of the path. Using the path for user ids should be done for resources (users, businesses, comments), but I don't think a map counts as one. We would want something like /map?sw_lat=val&sw_lng=val etc.

capstone/src/test/java/com/google/sps/servlets/MapServletTest.java
winniewan: Yes

capstone/src/main/java/com/google/sps/servlets/MapServlet.java
winniewan: I'm going to keep the user having a geoPt since they also have a mini google maps. We can use this information in the future. This filter will account for businesses only.

capstone/src/main/java/com/google/sps/servlets/MapServlet.java
winniewan: Yep

capstone/src/main/java/com/google/sps/servlets/MapServlet.java
winniewan: Yes, I agree! Thanks, done

capstone/src/main/java/com/google/sps/servlets/MapServlet.java
eash130: I guess I was kinda confused with the StContainsFilter and how that works. maybe you could add a comment describing what that filter is "filtering" for.

winniewan: Yep, definitely

capstone/src/main/java/com/google/sps/servlets/MapServlet.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet responsible for listing the businesses within the requested map radius. */
+@WebServlet("/map/*")
```
prabenzom, line 54: can we get rid of the "/*" portion?

winniewan: Yes, good catch!

PR #60, opened 2020-07-22
Fix POST request bug

winniewan:
The if statement include a return which skips over the rest of the function. This is not intended and the edit button will not work for business profiles. The fix is to this bug is to add the else statement.


alygerharter:
can you add a better description? it's not clear to me what the bug was.

PR #62, opened 2020-07-24
Update cloudbuild.yaml 

lukasberglund:
This fixes a bug that happened during deployment in which datastore was unable to load certain filterqueries.

### DIFF
--- a/deploy.yaml
+++ b/deploy.yaml
@@ -1,5 +1,9 @@
  steps:
  - name: maven:3-jdk-8
    entrypoint: mvn
+   args: ['appengine:deployIndex']
+   dir: 'capstone/'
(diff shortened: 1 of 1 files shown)


deploy.yaml
```diff
@@ -1,4 +1,8 @@
  steps:
+ - name: maven:3-jdk-8
+   entrypoint: mvn
+   args: ['appengine:deployIndex']
+   dir: 'capstone/'
  - name: maven:3-jdk-8
```
eash130, line 6: I'm not really familiar with maven, but would you need this next set of lines?

lukasberglund: Yeah. Both steps are necessary.

PR #63, opened 2020-07-24
Show timestamp on comments

lukasberglund:


capstone/src/main/java/com/google/sps/data/Comment.java
```diff
@@ -50,6 +53,7 @@ public Comment(
     this.name = name;
     this.businessId = businessId;
     this.parentId = parentId;
+    this.timestampStr = new SimpleDateFormat("MM/dd/yy HH:mm").format(timestamp);
```
eash130, line 56: So the epoch time is stored in datastore and then is just converted to a more readable string?

eash130: Do you need
this.timestamp = timestamp
then?

winniewan, line 56: Can you note which time stamp zone this gives?

alygerharter: The long timestamp is still needed if you intend to sort by timestamp.

eash130: Well yes, but the sort isn't occurring in the javascript right?

lukasberglund: That't a fair question. I am keeping the long in case the javascript still needs to access the raw timestamp.

PR #65, opened 2020-07-24
Design the view business page MVP with bootstrap

winniewan:


winniewan:
> Looks really nice. Kinda fits with Google's material design.
> 
> Could you change the color of the navbar to the red that you are using here?

PR #67, opened 2020-07-24
Display comments better on the users page

lukasberglund:
With this new pull request, comments on a users page are displayed with a wrapper that links to the business the comment was originally posted on.


capstone/src/main/webapp/js/util.js
```diff
@@ -57,17 +57,26 @@ export function setProfileUrl() {
...
+  linkElement.href = url;
+
+  return linkElement;
+}
+
 /** Helper function for building the login/logout link. */
 function buildLoginOrLogoutLink(url, isLoggedin) {    
-  const aElement = document.createElement('a');
-  aElement.setAttribute('href', url);
-  aElement.classList.add('nav-link');
```
winniewan, line 64: This class name is needed

lukasberglund: What do you mean?

PR #69, opened 2020-07-24
Let CommentServlet deal with hasReplies property

lukasberglund:
CommentServlet now initializes the hasReplies property as false initially, but updates the property if a reply is posted.

I also moved a function from `CommentsServletTest.java` to `CommentTestUtil.java` so that it can be used in `CommentServletTest.java`.


capstone/src/main/java/com/google/sps/servlets/comments/CommentServlet.java
```diff
@@ -61,13 +64,27 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
-      datastore.put(buildCommentEntity(request, currentUser.getUserId()));
-    } else {
+    if (currentUser == null) {
       response.sendError(
           HttpServletResponse.SC_UNAUTHORIZED, "User must be logged in to post comment");
+      return;
     }
 
+    String parentId = request.getParameter(PARENT_ID_PROPERTY);
+    if (parentId != null && !parentId.equals("")) {
```
eash130, line 74: When setting the parentId in doPost(), you can just default set parentId = "" if there is no parent so that you don't have to check for null and empty string.

capstone/src/main/java/com/google/sps/servlets/comments/CommentServlet.java
```diff
@@ -86,6 +103,7 @@ private Entity buildCommentEntity(HttpServletRequest request, String userId) {
     }
 
     commentEntity.setProperty(TIMESTAMP_PROPERTY, System.currentTimeMillis());
+    commentEntity.setProperty(HAS_REPLIES_PROPERTY, false);
```
eash130, line 106: Move this to inside the conditional in this method.

capstone/src/main/java/com/google/sps/servlets/comments/CommentServlet.java
lukasberglund: Can you show me the piece of code I would have to use to do that?

capstone/src/main/java/com/google/sps/servlets/comments/CommentServlet.java
lukasberglund: Doing that messes up my tests because I search for replies that have the `hasReplies` property set to false. Could I just stick with this? I mean it's a fact that a reply doesn't have any replies.

capstone/src/main/java/com/google/sps/servlets/comments/CommentServlet.java
lukasberglund: I changed it to just check if parentId != null which passed all tests so I think that also works.

alygerharter: but do you have a test for this specifically? :) reading the code, it looks like if parentId is empty you'll see some unexpected behavior, so better to code against that.

are you able to use Strings.isNullOrEmpty? https://guava.dev/releases/19.0/api/docs/com/google/common/base/Strings.html

capstone/src/main/java/com/google/sps/servlets/comments/CommentServlet.java
alygerharter: i think this is fine - the default for new comments is to have no replies.

capstone/src/test/java/com/google/sps/util/CommentTestUtil.java
```diff
@@ -0,0 +1,56 @@
...
+    comment.setProperty(USER_ID_PROPERTY, userId);
+    comment.setProperty(BUSINESS_ID_PROPERTY, businessId);
+    comment.setProperty(PARENT_ID_PROPERTY, parentId);
+    comment.setProperty(HAS_REPLIES_PROPERTY, false);
+
+    return comment;
+  }
+
+  public static String generateUniqueCommentId(long timestamp, String userId, String businessId) {
+    return "1" + timestamp + userId + businessId;
```
alygerharter, line 54: why "1"?

capstone/src/main/java/com/google/sps/servlets/comments/CommentServlet.java
alygerharter: never mind - according to the documentation, getParameter for a parameter that does not exist will always return null, not empty.

lukasberglund: I added a test for this case and made it so that it returns a specific error if the parentId is an empty string. I return an SC_BAD_REQUEST error with the text ""parentId cannot be an empty String. Either set a specific parentId or don't set the parameter at all." Is that good?

capstone/src/main/java/com/google/sps/servlets/comments/CommentServlet.java
```diff
@@ -61,13 +64,36 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
           HttpServletResponse.SC_UNAUTHORIZED, "User must be logged in to post comment");
+      return;
     }
 
+    String parentId = request.getParameter(PARENT_ID_PROPERTY);
+
+    if (parentId != null) {
+      if (parentId == "") {
+        response.sendError(
+            HttpServletResponse.SC_BAD_REQUEST,
```
alygerharter, line 78: sorry for the back and forth - rather than throwing an error here, i think just a check for null or empty on line 75 is sufficient.

either use Strings.isNullOrEmpty or check for an empty string (using parentId.isEmpty(), not ==)

capstone/src/test/java/com/google/sps/util/CommentTestUtil.java
lukasberglund: When I had it it was to make sure that if it's converted to a number, the leading zeros aren't dropped but I don't think that's necessary anymore so I removed it.

capstone/src/main/java/com/google/sps/servlets/comments/CommentServlet.java
eash130: yup I interpreted it wrong.

PR #70, opened 2020-07-24
Implement get functionality for FollowServlet

lukasberglund:
This feature is pretty simple. You use it to check whether you are currently following a business. You must be logged in to make this get request.


capstone/src/main/java/com/google/sps/servlets/follow/FollowServlet.java
```diff
@@ -156,4 +157,25 @@ public void doDelete(HttpServletRequest request, HttpServletResponse response)
...
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    User currentUser = userService.getCurrentUser();
+    if (currentUser == null) {
+      response.sendError(
+          HttpServletResponse.SC_UNAUTHORIZED, "User must be logged in to unfollow a business.");
+      return;
+    }
+
+    String businessId = request.getParameter(BUSINESS_ID_PROPERTY);
+    if (businessId == null) {
```
winniewan, line 171: Should you also check if the business id does not exist in datastore?

lukasberglund: I don't think that's necessary. If the business doesn't exist in the datastore the follow won't exist either so the servlet would just return `false` which is fine imo.

PR #73, opened 2020-07-27
Implement scrollable business side bar for the home page.

eash130:


capstone/src/main/webapp/css/index.css
```diff
@@ -1,106 +1,130 @@
...
+  -ms-flex: 1 1 190px;
+  flex: 1 1 190px;
+  margin: 0 8px;
 }
 
 #map {
-    height: 500px;
-    width: 700px;
+  height: 83vh;
+  width: 55vw;
```
eash130, line 91: I resorted to these units since I couldn't find another way to fit the screen properly.

PR #75, opened 2020-07-27
Change redirect link for new users to their profile page

winniewan:
This improves user flow when they create a new account, they should be redirected to their profile page to make changes.

### DIFF
--- a/capstone/src/main/java/com/google/sps/servlets/authentication/NewUserServlet.java
+++ b/capstone/src/main/java/com/google/sps/servlets/authentication/NewUserServlet.java
@@ -68,7 +68,7 @@ public class NewUserServlet extends HttpServlet {
 
       datastore.put(userEntity);
 
-      response.sendRedirect("/profile.html?id="+userId);
+      response.sendRedirect("/profile.html?id=" + userId);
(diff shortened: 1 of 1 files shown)


capstone/src/main/java/com/google/sps/servlets/authentication/NewUserServlet.java
```diff
@@ -68,6 +68,6 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
       datastore.put(userEntity);
     }
 
-    response.sendRedirect("/index.html");
+    response.sendRedirect("/profile.html?id=" + userId);
```
eash130, line 71: I think you can put response.sendRedirect("/index.html') in the try block and response.sendRedirect("/profile.html?id=" + userId) at the end of the catch block

winniewan: I don't think it's necessary since there's a try catch above

winniewan: yes! done.

PR #76, opened 2020-07-27
Add search servlets

eash130:


capstone/src/main/java/com/google/sps/servlets/business/BusinessServlet.java
```diff
@@ -141,6 +149,25 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+            .addField(
+                Field.newBuilder()
+                    .setName("name")
+                    .setTokenizedPrefix((String) businessEntity.getProperty(NAME_PROPERTY)))
+            .build();
+
+    try {
+      index.put(document);
+    } catch (PutException e) {
+      if (StatusCode.TRANSIENT_ERROR.equals(e.getOperationResult().getCode())) {
```
winniewan, line 166: Why are we still calling .put with the exception?

capstone/src/main/java/com/google/sps/servlets/business/BusinessServlet.java
```diff
@@ -141,6 +149,25 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+            .setId(id)
+            .addField(
+                Field.newBuilder()
+                    .setName("name")
+                    .setTokenizedPrefix((String) businessEntity.getProperty(NAME_PROPERTY)))
+            .build();
+
+    try {
+      index.put(document);
+    } catch (PutException e) {
```
winniewan, line 165: Should there be a return after the if statement?

capstone/src/main/java/com/google/sps/servlets/business/SearchServlet.java
```diff
@@ -0,0 +1,108 @@
...
+
+    List<BusinessProfile> businesses = new ArrayList<>();
+    try {
+      Results<ScoredDocument> searchResults =
+          index.search(
+              com.google.appengine.api.search.Query.newBuilder()
+                  .build("name:\"" + searchItem + "\""));
+
+      DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+      for (ScoredDocument document : searchResults) {
```
winniewan, line 70: I think this can be outside of the try statement.

capstone/src/main/java/com/google/sps/servlets/business/SearchServlet.java
```diff
@@ -0,0 +1,108 @@
...
+import com.google.sps.data.BusinessProfile;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet responsible for returning appropriate search results. */
```
winniewan, line 51: Can you add more to this comment, the option description you've sent in the group chat. It will give more clarity of the method you're using and what specific search this is.

eash130:
I will make a new commit adding comments to improve readability

capstone/src/main/java/com/google/sps/servlets/business/BusinessServlet.java
eash130: Nope, still need to put into datastore after try catch block

capstone/src/main/java/com/google/sps/servlets/business/BusinessServlet.java
eash130: The documentation calls for re-putting the document into the index if a transient error occurs.

capstone/src/main/java/com/google/sps/servlets/business/SearchServlet.java
eash130: I think it's fine to leave it in here, the documentation has it in the try and I'm not sure if there is an unforseen SearchException that might occur with accessing the ScoredDocuments in the loop.

winniewan: This looks like a huge try catch statement. You can test out which gives an exception and break it down from there. If there's an exception accessing each searchResults... then I guess there's no other way.

capstone/src/main/java/com/google/sps/servlets/business/SearchServlet.java
```diff
@@ -0,0 +1,108 @@
...
+        String location = (String) businessEntity.getProperty(LOCATION_PROPERTY);
+        String story = (String) businessEntity.getProperty(STORY_PROPERTY);
+        String about = (String) businessEntity.getProperty(ABOUT_PROPERTY);
+        String support = (String) businessEntity.getProperty(SUPPORT_PROPERTY);
+
+        BusinessProfile business =
+            new BusinessProfile(id, name, location, bio, story, about, email, support, false);
+        businesses.add(business);
+      }
+    } catch (SearchException e) {
```
winniewan, line 101: Can you add a test for when it enters this exception?

eash130: I'm not quite sure how to trigger that error. Lemme look into manually throwing an exception at a certain point in the try block.

eash130: Lemme look into how I might go about triggering a searchexception

alygerharter: I believe the SearchException comes from index.search()

eash130:
I added a test for SearchQueryException using escape characters in the search string.
I was unable to find a way to trigger SearchException since it depends on the SearchServiceFactory to be faulty. There also wasn't a way to mock it since it involves a non static method instantiated within doGet()

capstone/src/test/java/com/google/sps/servlets/business/SearchServletTest.java
```diff
@@ -0,0 +1,267 @@
...
+
+    Gson gson = new Gson();
+    JsonParser parser = new JsonParser();
+    Assert.assertEquals(parser.parse(servletResponse), parser.parse(gson.toJson(expectedResults)));
+  }
+
+  /** Test retrieving a business with an invalid search term. */
+  @Test
+  public void testDoGetSearchException() throws IOException {
+    doReturn("\"invalid").when(request).getParameter("searchItem");
```
alygerharter, line 262: so what happens if the business name has a special character? what if i want to search for "eashan's pizza"?

capstone/src/test/java/com/google/sps/servlets/business/SearchServletTest.java
```diff
@@ -0,0 +1,267 @@
...
+    } catch (PutException e) {
+      if (StatusCode.TRANSIENT_ERROR.equals(e.getOperationResult().getCode())) {
+        index.put(document);
+      }
+    }
+  }
+
+  /** Test retrieving a business by exact match to business name. */
+  @Test
+  public void testDoGetExactMatch() throws IOException {
```
alygerharter, line 146: please add a test case for when there are no matches.

capstone/src/test/java/com/google/sps/servlets/business/SearchServletTest.java
eash130: A SearchException is only raised for the \" and \\ escape characters. All other escape characters work, so "eashan's pizza" is a valid Search Query.

alygerharter: okay, please add a test for that as well.

eash130: See latest commit, it provides a test case for it, tough I'm not sure if that was the best way to do it.

alygerharter: i still don't see a test for that..? i only see the invalid character test.

what i'm looking for is a test where the business is named something like "eashan's pizza" and the search term is "eashan's".

eash130: Gotcha, I just added it in.

PR #78, opened 2020-07-28
Fix minor bug in comments.js

lukasberglund:
The bug was due to the fact that we were assigning `commentContainer` without declaring it.


capstone/src/main/webapp/js/comments.js
```diff
@@ -65,7 +65,7 @@ function buildHiddenStaticFormField(name, value) {
 
 /** Load a list of comments that the user posted */
 export function loadUserCommentList(userId) {
-  commentContainer = document.createElement('div');
+  const commentContainer = document.createElement('div');
```
winniewan, line 68: Probably add a short description of the minor bug

PR #80, opened 2020-07-28
Fix uncaught bugs in Comment and Comments Servlet

lukasberglund:
The first bug was that when a name wasn't speficied for a comment but instead a random id was generated, the Comment Servlet was not able to retrieve comments to update their `hasReplies` property. 

The bug wasn't caught because in order to reference comments more easily in testing, we were manually assigning names to them. This was changed in CommentServletTest.java.

The second bug was that when a comment was generated the `hasReplies` property was always set to `false` even if it was `true`. 

This bug wasn't caught because the `generateTestComment` function ignored the `hasReplies` field and would just generate a test comment with the `hasReplies` field always set to zero, so the expected results matched the actual results because they were both wrong in the same way.

### DIFF
--- a/capstone/src/main/java/com/google/sps/data/Comment.java
+++ b/capstone/src/main/java/com/google/sps/data/Comment.java
@@ -39,33 +39,14 @@ public final class Comment {
   private final boolean hasReplies;
   private final String timestampStr;
 
-  public Comment(
+  public Comment (
       String id,
       String content,
       long timestamp,
       String userId,
       String name,
       String businessId,
-      String parentId) {
-    this.id = id;
-    this.content = content;
-    this.timestamp = timestamp;
-    this.userId = userId;
-    this.name = name;
-    this.businessId = businessId;
-    this.parentId = parentId;
-    this.hasReplies = false;
-    // Epoch timestamp is formatted in UTC time
-    this.timestampStr = new SimpleDateFormat("MM/dd/yy HH:mm").format(timestamp);
-  }
-
-  public Comment(
-      String id,
-      String content,
-      long timestamp,
-      String userId,
-      String name,
-      String businessId,
+      String parentId,
       boolean hasReplies) {
     this.id = id;
     this.content = content;
@@ -73,7 +54,7 @@ public final class Comment {
     this.userId = userId;
     this.name = name;
     this.businessId = businessId;
-    this.parentId = "";
+    this.parentId = parentId;
     this.hasReplies = hasReplies;
     // Epoch timestamp is formatted in UTC time
     this.timestampStr = new SimpleDateFormat("MM/dd/yy HH:mm").format(timestamp);
--- a/capstone/src/main/java/com/google/sps/data/Comment.java
+++ b/capstone/src/main/java/com/google/sps/data/Comment.java
@@ -74,7 +74,7 @@ public final class Comment {
     this.name = name;
     this.businessId = businessId;
     this.parentId = "";
-    this.hasReplies = false;
+    this.hasReplies = hasReplies;
     // Epoch timestamp is formatted in UTC time
     this.timestampStr = new SimpleDateFormat("MM/dd/yy HH:mm").format(timestamp);
   }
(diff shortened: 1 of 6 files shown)


capstone/src/main/java/com/google/sps/data/Comment.java
```diff
@@ -74,7 +74,7 @@ public Comment(
     this.name = name;
     this.businessId = businessId;
     this.parentId = "";
-    this.hasReplies = false;
+    this.hasReplies = hasReplies;
```
alygerharter, line 77: to avoid some further confusion, i'd just make one constructor that takes both parentId and hasReplies. you can throw an IllegalArgumentException for a bad/impossible combination.

capstone/src/main/java/com/google/sps/util/CommentDatastoreUtil.java
```diff
@@ -43,8 +43,13 @@ public static Comment generateComment(Entity commentEntity, DatastoreService dat
...
     String parentId = (String) commentEntity.getProperty(PARENT_ID_PROPERTY);
-    boolean hasReplies = (boolean) commentEntity.getProperty(HAS_REPLIES_PROPERTY);
 
-    return new Comment(id, content, timestamp, userId, name, businessId, parentId);
+    // Use two different constructors depending on whether the comment is a reply or not
+    if (parentId.isEmpty()) {
+      boolean hasReplies = (boolean) commentEntity.getProperty(HAS_REPLIES_PROPERTY);
+      return new Comment(id, content, timestamp, userId, name, businessId, hasReplies);
+    } else {
+      return new Comment(id, content, timestamp, userId, name, businessId, parentId);
```
alygerharter, line 52: if you get rid of the extra constructor, you can also get rid of this if/else.

capstone/src/main/java/com/google/sps/data/Comment.java
lukasberglund: Since it's just a constructor meant for displaying JSON, would it be okay not to check for bad combinations?

lukasberglund: Do you think I should throw an IllegalArgumentException or is it okay to just make sure that doesn't happen elsewhere?

lukasberglund: Done!

alygerharter: better to do the enforcement here

capstone/src/test/java/com/google/sps/servlets/comments/CommentsServletTest.java
```diff
@@ -160,14 +161,29 @@ private Comment generateCommentForTest(
...
     String id = generateUniqueCommentId(timestamp, userId, businessId);
     return new Comment(
-        id, id, timestamp, userId, getProfileName(userId, ds), businessId, hasReplies);
+        id,
+        id,
+        timestamp,
+        userId,
+        getProfileName(userId, ds),
+        businessId, /*parentId*/
+        "",
```
winniewan, line 170: You can set this as a variable, EMPTY_ID or similar.

capstone/src/test/java/com/google/sps/servlets/comments/CommentsServletTest.java
```diff
@@ -160,14 +161,29 @@ private Comment generateCommentForTest(
...
-    return new Comment(id, id, timestamp, userId, getProfileName(userId, ds), businessId, parentId);
+    return new Comment(
+        id,
+        id,
+        timestamp,
+        userId,
+        getProfileName(userId, ds),
+        businessId,
+        parentId, /*hasReplies*/
+        false);
```
winniewan, line 186: You can set a variable, NO_REPLIES, or similar

lukasberglund: Isn't that a bit overdone? I mean a boolean has to be exactly one of two values and it's clear which is which in this case.

winniewan: I think it's fine since you have the comment there as well

PR #88, opened 2020-07-30
Add a working follow button to business pages 

lukasberglund:


capstone/src/main/webapp/js/util.js
```diff
@@ -108,11 +108,21 @@ export function getJsonObject(url, parameters = {}) {
...
+  
+  return fetch(url, {method: type});
+}
+
 export function makePostRequest(url, parameters) {
-  return fetch('https://example.com/profile', {
-    method: 'POST',
-    body: JSON.stringify(parameters),
-  });
+  return makeRequest(url, parameters, 'POST');
```
alygerharter, line 125: no need to add an extra nested function. just put the contents of makeRequest here.

lukasberglund: Why is that better form?

alygerharter: you aren't reusing makeRequest() anywhere, so then you just have a function around a function.  it's not really adding anything, here.

lukasberglund: I do use it in `business.js` on line 203. If you want I can get rid of makePostRequest and replace those uses with `makeRequest(url, parameters, 'POST');` though.

lukasberglund: Update: I removed `makePostRequest()`

alygerharter: this is cleaner, thanks!

PR #90, opened 2020-07-30
Fix login and logout functionality due to recent changes with internal use restrictions

winniewan:


capstone/src/test/java/com/google/sps/servlets/authentication/LoginServletTest.java
```diff
@@ -56,7 +56,7 @@
   private static final String EMAIL = "abc@gmail.com";
   private static final String AUTHDOMAIN = "gmail.com";
   private static final String LOG_IN_URL = "/_ah/login?continue=%2Fcheck_new_user";
-  private static final String LOG_OUT_URL = "/logout";
+  private static final String LOG_OUT_URL = "/_ah/logout?continue=index.html";
```
eash130, line 59: just curious, why does this work?

winniewan: This is just the original logout button. 
I had a separate servlet with deleting cookies, but that won't work after the restriction is placed since you're required to login to use our webpage.

PR #97, opened 2020-08-03
Implement single calendar for followed businesses

eash130:


eash130:
Not optimal solution, working on better JS for fetching.

PR #101, opened 2020-08-04
Fix issue with show more not working on search results.

eash130:
Originally, search results were not able to have the about description expanded since there are multiple ids. Even though the other two divs were hidden, the ids were still "active" and those were being read first. By moving the results div first, it would ensure that these ids are read first.


lukasberglund:
Can you explain how this fixes the issue?

